PR #8, opened 2023-02-06
Add basic CI tools

kajinamit:
This introduces tools as well as definitions to run basic CI jobs following the other operators.

Signed-off-by: Takashi Kajinami <tkajinam@redhat.com>

### DIFF
--- a/.github/create_bundle.sh
+++ b/.github/create_bundle.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+set -e
+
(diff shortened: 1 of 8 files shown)


bshephar:
I wonder if we would be better off waiting until this is closer to working properly before enabling CI? Things like go lint, go fat would be fine. But I haven't looked at the Dockerfile too much yet, or any unit tests. So some of the things here might cause headaches if we want to arbitrarily merge changes during the initial build and collaboration phase?

PR #12, opened 2023-02-07
Use k8s secret for Horizon secret_key

bshephar:


controllers/horizon_controller.go
```diff
@@ -471,6 +471,22 @@ func (r *HorizonReconciler) generateServiceConfigMaps(
 		return nil
 	}
 
+	tmpl := []util.Template{
+		{
+			Name:       horizon.ServiceName,
+			Namespace:  instance.Namespace,
+			Type:       util.TemplateTypeNone,
+			CustomData: map[string]string{"horizon-secret": instance.Spec.HorizonSecret},
```
kajinamit, line 479: I think this should not be part of spec and there are a few options

- Accept secret name in spec and use it to look up the secret
- Use the hard-coded secret name and generate the secret if it does not exist. This allows users to deploy horizon without creating secrets, and users can override it by modfiying/pre-creating the secret
- Add secret key to global secret instance and look it up.

PR #14, opened 2023-02-07
Drop secret from spec

kajinamit:
The SECRET_KEY parameter is sensitive information and should not be part of spec. Because the value is internal and is not exposed to clients, this makes the value auto-generated without user input.

### DIFF
--- a/api/v1alpha1/horizon_types.go
+++ b/api/v1alpha1/horizon_types.go
@@ -77,10 +77,6 @@ type HorizonSpec struct {
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default:"HorizonSecretKey"
-	HorizonSecret string `json:"horizonSecret,omitempty"`
-
-	// +kubebuilder:validation:Optional
 	// HorizonRoute holds all of the necessary options for configuring the Horizon Route object.
 	// This can be used to configure TLS
 	//TODO(bshephar) Implement everything about this. It's just a placeholder at the moment.
--- a/config/crd/bases/horizon.openstack.org_horizons.yaml
+++ b/config/crd/bases/horizon.openstack.org_horizons.yaml
@@ -65,8 +65,6 @@ spec:
                   to add additional files. Those get added to the service config dir
                   in /etc/<service> . TODO: -> implement'
                 type: object
-              horizonSecret:
-                type: string
               nodeSelector:
                 additionalProperties:
                   type: string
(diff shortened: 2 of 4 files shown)


pkg/horizon/secret.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"encoding/base64"
+
+	"math/rand"
+	"time"
+)
+
+// GenerateSecret -
+func GenerateSecret() string {
```
stuggi, line 26: it might make sense to add this to lib-common and pass in the pwd length? a while back we were talking about create service passwords if none were passed in via a secret which match the service selector. so this could be used there as well I think.

pkg/horizon/secret.go
```diff
@@ -0,0 +1,33 @@
...
+	rand.Seed(time.Now().UnixNano())
+	data := make([]byte, 10)
+	for i := 0; i < 10; i++ {
+		data[i] = byte(rand.Intn(10))
+	}
+	return base64.StdEncoding.EncodeToString(data)
```
bshephar, line 32: Does it make sense to base64 encode it here when any string we push into the secret will be b64 encoded anyway? For example, the current blahsecret:

```
❯ oc get secret -o jsonpath='{.data}' horizon | jq
{
  "horizon-secret": "YmxhaFNlY3JldA=="
}
```

kajinamit: I picked up this implementation from keystone operator (the code was used to generate fernet keys) but I noticed we don't need base64 encode here. will make this more simple.

pkg/horizon/secret.go
kajinamit: +1. Let me work on the implementation in lib-common.

kajinamit: I've pushed https://github.com/openstack-k8s-operators/lib-common/pull/172 .

kajinamit: As is suggested in the above PR I'm trying the rand.String from apimachinery. Let me hold this PR until I test this out.

kajinamit:
Verified this change in my local env.

```
$ oc describe secrets/horizon
Name:         horizon
Namespace:    openstack
Labels:       horizon.openstack.org/name=horizon
              horizon.openstack.org/namespace=openstack
              horizon.openstack.org/uid=65dc80f0-29d4-4f01-8a7f-0c8153039a67
Annotations:  <none>

Type:  Opaque

Data
====
horizon-secret:  10 bytes

$ oc exec -t horizon-5bff8d5fd5-x9nbn -- cat /etc/openstack-dashboard/.horizon-secret
mlmjj5g47k
```

bshephar (approved):
Yeah, looks much better this way. Nice one

PR #35, opened 2023-02-21
Drop unnecessary lines from Dockerfile

kajinamit:
/workspace does not exist in the container image and building the image fails with the following error.

```
[2/2] STEP 13/19: COPY --from=builder /workspace/manager .
Error: building at STEP "COPY --from=builder /workspace/manager .": checking on sources under "...": copier: stat: "/workspace/manager": no such file or directory
```

The file should be copied into $DEST_ROOT instead.

Also, user is already twice and the second definition looks redundant.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -60,8 +60,6 @@ ENV USER_UID=$USER_ID \
     OPERATOR_TEMPLATES=/usr/share/horizon-operator/templates/
 
 WORKDIR /
-COPY --from=builder /workspace/manager .
-USER 65532:65532
 
 # Install operator binary to WORKDIR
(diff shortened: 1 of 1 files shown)


kajinamit:
I'm using the following steps to try building images but I might be missing something fundamental.
```
$ export VERSION=0.0.1
$ export IMAGE_TAG_BASE=default-route-openshift-image-registry.apps-crc.testing/openstack/horizon-operator
$  export IMG=$IMAGE_TAG_BASE:v$VERSION
$ make build docker-build docker-push bundle bundle-build bundle-push catalog-build catalog-push
```

bshephar (approved):

PR #39, opened 2023-02-22
Drop namespace from LeaderElectionID

kajinamit:
The LeaderElectionID should not contain '/', which causes an error while acquiring leader lock.


bshephar (approved):
Looks better as openstack.org. Thanks again.

PR #41, opened 2023-02-22
Add missing rbac to access keystone CRs

kajinamit:
Currently horizon operator requires access to keystone CRs to obtain url of keystone endpoint. This adds missing RBAC to allow that access.


kajinamit:
Note to self: Because horizon does not require its own user it does not require access to keystoneservices CR.

bshephar (approved):
lgtm, again, thanks.

PR #42, opened 2023-02-22
Use separate service account for deployment

kajinamit:
... instead of using the operator service account. This follows the current standard in oko.


bshephar:
lgtm thanks

PR #44, opened 2023-02-27
Add functional tests

bshephar:


bshephar:
There are two possible solutions to the problem here.

One, is that we follow the pattern outlined by the openstack-operator and comment out all of the content in `suite_test.go`:
https://github.com/openstack-k8s-operators/openstack-operator/blob/master/controllers/core/suite_test.go#L17

The second option is to ensure the envtest variables are defined in the CI job. Which would involve essentially setting up the options defined in the second commit on this PR:
https://github.com/openstack-k8s-operators/horizon-operator/pull/44/commits/0d9faef7c09d1a1d93102c53c6d7f304cd214285

https://github.com/openstack-k8s-operators/horizon-operator/blob/main/.github/workflows/golangci-lint.yaml#L26

Pending agreement on the approach that should be applied here. I guess my preference is for the second option and using the envtest setup, which I believe is more in-line with the testing standards outlined by Kubebuilder:
https://kubebuilder.io/cronjob-tutorial/writing-tests.html

.github/workflows/golangci-lint.yaml
```diff
@@ -22,8 +22,8 @@ jobs:
         run: ./openstack-k8s-operators-ci/test-runner/govet.sh
       - name: Run golint.sh
         run: ./openstack-k8s-operators-ci/test-runner/golint.sh
-      - name: Run gotest.sh
-        run: ./openstack-k8s-operators-ci/test-runner/gotest.sh
+      - name: Run gotest
+        run: make gotest
```
gibizer, line 26: eventually this whole job should be moved to prow. But until that this is an easy way to run the tests.

Makefile
```diff
@@ -103,7 +103,7 @@ vet: ## Run go vet against code.
 
 .PHONY: test
 test: manifests generate fmt vet envtest ## Run tests.
-	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out
+	OPERATOR_TEMPLATES=./templates KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test ./... -coverprofile cover.out
```
gibizer, line 106: setting the OPERATOR_TEMPLATE here is actually better than what we do in nova setting in programmatically in each test.  https://github.com/openstack-k8s-operators/nova-operator/blob/e6742035a36d7c865620511d25d5784feb721999/test/functional/nova_controller_test.go#L61

Makefile
```diff
@@ -289,9 +289,9 @@ govet: get-ci-tools
 	GOWORK=off $(CI_TOOLS_REPO_DIR)/test-runner/govet.sh ./api
 
 # Run go test against code
-gotest: get-ci-tools
-	GOWORK=off $(CI_TOOLS_REPO_DIR)/test-runner/gotest.sh
-	GOWORK=off $(CI_TOOLS_REPO_DIR)/test-runner/gotest.sh ./api
+gotest: envtest get-ci-tools
+	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" GOWORK=off $(CI_TOOLS_REPO_DIR)/test-runner/gotest.sh
```
gibizer, line 293: I'm not a fan of the scripts in the CI_TOOLS_REPO as I have not much control over it so I try to avoid them. In nova we ended up basically making gotest an alias for the test target. And having the test target run every tests. If we ever need both functional envtest and simple unit test then we might need to split that target to two but so far that worked OK for us.

test/functional/horizon_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+	"k8s.io/apimachinery/pkg/types"
+
+	horizon "github.com/openstack-k8s-operators/horizon-operator/api/v1alpha1"
+	keystonev1 "github.com/openstack-k8s-operators/keystone-operator/api/v1beta1"
+	condition "github.com/openstack-k8s-operators/lib-common/modules/common/condition"
+)
+
+const (
+	// TODO(gibi): Do we want to test in a realistic namespace like "openstack"?
+	TestNamespace = "default"
```
gibizer, line 23: one hint. In envtest you cannot really delete a namespace https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation but probably you want to isolate each test case from each other so if one leaves some resources in the namespace it does not effect the later test running in that namespace. So we started using a generated namespace name for each test to isolate them.

test/functional/horizon_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+		Template:  template,
+		Instance:  &horizon.Horizon{},
+	}
+}
+
+// Creates the Horizon resource in k8s based on the Template. The Template
+// is not updated during create. This call waits until the resource is created.
+// The last known state of the resource is available via Instance.
+func (t TestHorizon) Create() {
+	Expect(k8sClient.Create(ctx, t.Template.DeepCopy())).Should(Succeed())
```
gibizer, line 72: One thing I learned in the hard way was that using golang structs for creating resources has a non trivial side effect. When you say horizon.Horizon{} you get a full go struct with all the fields have some a value. If you don't provide a field in the template then go will initialize it with the null value regardless if you have kubebilder annotations with default values. Then Create here will serialize the go stuct to json. This way you cannot really simulate that some field is not provided in the create request to let the API server do the defaulting during deserialization of the json payload. So in nova we switched to unstructured Create to be able to test with poper default values.  https://github.com/openstack-k8s-operators/nova-operator/blob/e6742035a36d7c865620511d25d5784feb721999/test/functional/base_test.go#L52-L86

test/functional/horizon_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+		// will fail to generate the ConfigMap as it does not find common.sh
+		err := os.Setenv("OPERATOR_TEMPLATES", "../../templates")
+		Expect(err).NotTo(HaveOccurred())
+
+		horizon = NewTestHorizon(TestNamespace)
+		horizon.Create()
+
+	})
+
+	AfterEach(func() {
```
gibizer, line 191: DeferCleanup is suggested by the ginkgo doc for cleanup as that can be set up close to the creation of the resource that needs cleanup

test/functional/horizon_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+
+	var horizon TestHorizon
+	var secret *corev1.Secret
+	var keystoneAPI *TestKeystoneAPI
+
+	BeforeEach(func() {
+		// lib-common uses OPERATOR_TEMPLATES env var to locate the "templates"
+		// directory of the operator. We need to set them othervise lib-common
+		// will fail to generate the ConfigMap as it does not find common.sh
+		err := os.Setenv("OPERATOR_TEMPLATES", "../../templates")
```
gibizer, line 183: you don't need this if you add it this env var to gotest make target as well similarly how you added it to the make test target

test/functional/horizon_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+		if keystoneAPI != nil {
+			keystoneAPI.Delete()
+		}
+	})
+
+	When("A Horizon instance is created", func() {
+
+		It("should have the Spec and Status fields initialized", func() {
+			Expect(horizon.Instance.Spec.Secret).Should(Equal("osp-secret"))
+			// TODO(gibi): Why defaulting does not work?
```
gibizer, line 205: the answer is the raw create above

gibizer (commented):
Thanks for start providing test coverage with envtest. 

I see you looked at the placement-operator as example. I suggest to look at the nova-operator as well as the coverage in placement-operator is mostly from last summer and since the we learned couple of tricks when created the coverage for the nova-operator. We have the intention to go back to placement-operator and extend the coverage there and probably synchronize the styles between the two.

gibizer:
Why this was closed? Is it superseded by another PR?

bshephar:
> Why this was closed? Is it superseded by another PR?

Not sure actually. Looks like I closed it and then pushed some more changes to it. Not too sure what happened there, I'll see if there is a reopen here somewhere.

gibizer:
For me the Reopen and comment button is grayed out. :/

PR #45, opened 2023-02-27
Support custom config

kajinamit:
Horizon supports loading multiple config files in the local_settings.d directory. This uses that mechanism to allow users to inject/override options.


kajinamit:
Verified in my local env
```
$ oc exec -it horizon-6c4bdbf789-pdbbd -- ls /etc/openstack-dashboard/local_settings.d
_90_manila_shares.py  9999_custom_settings.py
$ oc exec -it horizon-6c4bdbf789-pdbbd -- cat /etc/openstack-dashboard/local_settings.d/9999_custom_settings.py
SESSION_TIMEOUT = 3600
```

bshephar (approved):

PR #52, opened 2023-03-01
Add memcached

bshephar:


controllers/horizon_controller.go
```diff
@@ -547,3 +572,52 @@ func (r *HorizonReconciler) ensureHorizonSecret(
...
+}
+
+func (r *HorizonReconciler) checkMemcachedCR(ctx context.Context, instance *horizonv1alpha1.Horizon) (bool, error) {
+	list := &memcachedv1.MemcachedList{}
+	if err := r.Client.List(ctx, list); err != nil {
+		r.Log.Info("Error while getting list of memcaches")
+		return false, err
+	}
+	for _, memcached := range list.Items {
+		if memcached.ObjectMeta.Name == fmt.Sprintf("%s-memcached", instance.Name) {
```
kajinamit, line 599: the instance name should be configurable so that users can use a single cluster rather than multiple ones.

controllers/horizon_controller.go
```diff
@@ -340,6 +344,21 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
 		return ctrl.Result{}, err
 	}
 
+	memcachedExists, err := r.checkMemcachedCR(ctx, instance)
```
kajinamit, line 347: I think this approach makes sense instead of always creating/patching the CR, because 
- this allows users to pre-create memcached CR
- this allows users to customize memcache CR(like scaling out) after it is auto-created
while this means we can't apply any update for auto-created memcached CR during update.

Alternatively we can create or patch memached cr unless user gives memcached cr name, and create as many replicas as horizon pod, but I don't know this can be a better approach than this.

controllers/horizon_controller.go
```diff
@@ -547,3 +572,52 @@ func (r *HorizonReconciler) ensureHorizonSecret(
...
+	for _, memcached := range list.Items {
+		if memcached.ObjectMeta.Name == fmt.Sprintf("%s-memcached", instance.Name) {
+			r.Log.Info("Found Memcached CR for Horizon!")
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (r *HorizonReconciler) checkMemcachedSvc(instance *horizonv1alpha1.Horizon, h *helper.Helper) (svc string, err error) {
```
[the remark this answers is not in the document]
bshephar: Will update.

controllers/horizon_controller.go
bshephar: I wonder if we would want the user to specify the memcached instance. 

Or rather, just a Spec item like `sharedMemcached` that takes a bool. Then have a function that searches for existing `memcached` implementations.

I kind of prefer the latter

gibizer: I would like to make the connection between the Horizon/my-horizon instance and the Memcached/my-horizon-memcached explicit. As this is now hardcodes a connection implicitly that is pretty hard to discover from outside, looking at the CRs in the system.

I would do that by adding a MemcachedInstance field to HorizonSpec to carry the name of the Memcached CR instance the Horizon instance supposed to use.

I would make that field Optional. And then Horizon can create its own Memcached instance if value for the MemcachedInstance is not provided in the CR.

controllers/horizon_controller.go
```diff
@@ -340,6 +344,23 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
 		return ctrl.Result{}, err
 	}
 
+	if instance.Spec.SharedMemcached == "" {
+		memcachedExists, err := r.getMemcachedCR(ctx, fmt.Sprintf("%s-memcached", instance.Name))
+		if err != nil {
+			r.Log.Info("Failed retrieving Memcached CR list")
+			return ctrl.Result{}, err
+		}
```
gibizer, line 352: If you know that SharedMemcached is not provided in the CR then you don't need to try to look up the Memcached either and you can move ahead and create your own.

controllers/horizon_controller.go
```diff
@@ -340,6 +344,23 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
...
+		if !memcachedExists {
+			memcached := r.renderMemcached(instance)
+			err = r.Client.Create(ctx, memcached)
```
gibizer, line 355: I suggest to use the pattern of CreateOrPatch instead of trying to Get and if not exists Create the CR. CreateOrPatch will help also in case only some part of the Memcached instance needs to be updated (e.g. when instance.Spec.Replicas, changes). See for example https://github.com/openstack-k8s-operators/lib-common/blob/e77d8d19c2ffab69535fbbd47531b833d82fee49/modules/common/deployment/deployment.go#L57

controllers/horizon_controller.go
```diff
@@ -547,3 +579,52 @@ func (r *HorizonReconciler) ensureHorizonSecret(
...
 }
+
+func (r *HorizonReconciler) renderMemcached(instance *horizonv1alpha1.Horizon) *memcachedv1.Memcached {
+	return &memcachedv1.Memcached{
+		TypeMeta: metav1.TypeMeta{
+			APIVersion: "memcached.openstack.org/v1beta1",
+			Kind:       "Memcached",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-memcached", instance.Name),
```
gibizer, line 590: we probably not need to postfix the name of each Memcached instance with "-memcached" the Name only need to be unique per namespace per Kind.

controllers/horizon_controller.go
```diff
@@ -463,12 +485,22 @@ func (r *HorizonReconciler) generateServiceConfigMaps(
 		return err
 	}
 
+	if instance.Spec.SharedMemcached == "" {
+		memcachedSvc, err = r.checkMemcachedSvc(fmt.Sprintf("%s-memcached", instance.Name), h)
+		if err != nil {
+			return err
```
gibizer, line 491: This probably an case you want to reflect in the Status.Conditions.

controllers/horizon_controller.go
```diff
@@ -547,3 +579,52 @@ func (r *HorizonReconciler) ensureHorizonSecret(
...
+		r.Log.Info("Error getting podlist: %s", err)
+		return "", err
+	}
+	for _, service := range svcList.Items {
+		if service.Name == memcachedName {
+			return service.Name, nil
+		}
+	}
+	r.Log.Info("No Memcached service was found")
+	return "", nil
```
gibizer, line 629: as far as I see the usage of checkMemcachedSvc relies on getting the name of the service or getting an error. So you might want to return an error here.

controllers/horizon_controller.go
```diff
@@ -547,3 +579,52 @@ func (r *HorizonReconciler) ensureHorizonSecret(
...
+			r.Log.Info("Found Memcached CR for Horizon!")
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func (r *HorizonReconciler) checkMemcachedSvc(memcachedName string, h *helper.Helper) (svc string, err error) {
+	labelSelector := k8s_labels.Set{"app": "memcached"}.String()
+	svcList, err := h.GetKClient().CoreV1().Services("").List(context.TODO(), metav1.ListOptions{
```
gibizer, line 616: I think we have a lib-common helper for this https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/service/service.go#L218

controllers/horizon_controller.go
bshephar: It needs to have a different name compared to the instance.Name though, so I figured this makes sense. If I try to call it instance.Name, then the memcached CR returns the following error:
```
❯ oc get memcached horizon -o jsonpath='{.status.conditions[3].message}'
Service config create error occured error create/updating configmap: Object openstack/horizon-config-data is already owned by another Horizon controller horizon
```

controllers/horizon_controller.go
```diff
@@ -340,6 +344,29 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
 		return ctrl.Result{}, err
 	}
 
+	// Create Memcached instance if no shared instance exists.
+	if instance.Spec.SharedMemcached == "" {
+		memcached := r.renderMemcached(instance)
+		op, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), memcached, func() error {
+			memcached.Labels = map[string]string{"service": instance.Name}
+			err := controllerutil.SetControllerReference(helper.GetBeforeObject(), memcached, helper.GetScheme())
```
bshephar, line 352: Adding a label here is probably unnecessary since we have the owner reference set as well. Fairly easy to tell which service has created the memcached instance by just checking the `ownerReferences`:

```
❯ oc get memcached horizon-memcached -o jsonpath='{.metadata.ownerReferences}' | jq
[
  {
    "apiVersion": "horizon.openstack.org/v1alpha1",
    "blockOwnerDeletion": true,
    "controller": true,
    "kind": "Horizon",
    "name": "horizon",
    "uid": "48addfa8-3e8b-44f8-8d41-388e1eb8d00f"
  }
]
```

bshephar: Leaving it because it's something that was raised while discussing the approach for this. And there may be some established standard to lookup CR's by the label. So I'll take wider feedback on the topic.

controllers/horizon_controller.go
bshephar: This has been implemented in the latest patchset.

controllers/horizon_controller.go
bshephar: Using `controllerutil.CreateOrPatch()` now in the latest update.

controllers/horizon_controller.go
bshephar: I believe the latest update resolves this concern

controllers/horizon_controller.go
bshephar: Done in the latest update.

controllers/horizon_controller.go
bshephar: Implemented this in the latest update too. Thanks for the pointer.

controllers/horizon_controller.go
gibizer: Hm probably Memcached reconciler generates its own config map name based on just the instance.Name + "-config-data". I would say this is a problem in the Memcached implementation. I think Memcached should generate its config map name based on instance.Kind + instance.Name + "-config-data". That way we could avoid the need to each creator of a Memcached add the Kind to the name of the Memcached instance.

controllers/horizon_controller.go
```diff
@@ -340,6 +345,29 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
 		return ctrl.Result{}, err
 	}
 
+	// Create Memcached instance if no shared instance exists.
+	if instance.Spec.SharedMemcached == "" {
+		memcached := r.renderMemcached(instance)
+		op, err := controllerutil.CreateOrPatch(ctx, helper.GetClient(), memcached, func() error {
+			memcached.Labels = map[string]string{"service": instance.Name}
```
gibizer, line 352: // CreateOrPatch creates or patches the given object in the Kubernetes
// cluster. The object's desired state must be reconciled with the before
// state inside the passed in callback MutateFn.

bshephar: I guess I should be doing `memcached := memcachedv1.Memcached{}`. Rather than using that `renderMemcached` function in this scenario. Then set each of the fields within the `MutateFn()`.

I'll change it tomorrow and update accordingly.

controllers/horizon_controller.go
bshephar: Yeah, fair point. Rather than the Info log, it's better to return it as an error. This has been updated in the latest commit.

bshephar:
> 

ack yeah. I'll take that as a discussion point with the team that implemented memcached to come up with a suitable plan for them. I'm happy to loop back and update this once it's resolved there.

controllers/horizon_controller.go
```diff
@@ -308,6 +313,36 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
...
+			return ctrl.Result{}, err
+		}
+		if op != controllerutil.OperationResultNone {
+			r.Log.Info(fmt.Sprintf("Memcached %s - %s", memcached.Name, op))
+		}
+
+		// Mark the Memcached Service as Ready if we get to this point with no errors
+		instance.Status.Conditions.MarkTrue(
+			horizonv1alpha1.HorizonMemcachedReadyCondition, horizonv1alpha1.HorizonMemcachedReadyMessage)
+	}
```
kajinamit, line 344: (Can be follow-up)
Should we ensure we delete the default memcached CR in case SharredMemcache id given ?

controllers/horizon_controller.go
```diff
@@ -547,3 +608,32 @@ func (r *HorizonReconciler) ensureHorizonSecret(
 
 	return nil
 }
+
+func (r *HorizonReconciler) renderMemcached(instance *horizonv1alpha1.Horizon) *memcachedv1.Memcached {
```
kajinamit, line 612: Eventually this and checkMemcachedSvc shuld be part of memcached operator, I think

templates/horizon/config/local_settings.py
```diff
@@ -94,12 +94,12 @@
...
 # https://docs.djangoproject.com/en/1.11/topics/http/sessions/.
-#CACHES = {
-#    'default': {
-#        'BACKEND': 'django.core.cache.backends.memcached.MemcachedCache',
-#        'LOCATION': '127.0.0.1:11211',
-#    },
-#}
+CACHES = {
+    'default': {
+        'BACKEND': 'django.core.cache.backends.memcached.PyMemcacheCache',
```
kajinamit, line 99: this is good as MemcachedCache has been deprecated

templates/horizon/config/local_settings.py
```diff
@@ -94,12 +94,12 @@
...
+        'LOCATION': '{{.memcachedSvc}}:11211',
```
kajinamit, line 100: PyMemcacheCache does not require inet6 prefix so this should work for both IPv4 and IPv6.

controllers/horizon_controller.go
```diff
@@ -547,3 +608,32 @@ func (r *HorizonReconciler) ensureHorizonSecret(
...
+			Kind:       "Memcached",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      fmt.Sprintf("%s-memcached", instance.Name),
+			Namespace: instance.Namespace,
+		},
+	}
+}
+
+func (r *HorizonReconciler) checkMemcachedSvc(ctx context.Context, memcachedName string, h *helper.Helper, namespace string) (svc string, err error) {
```
kajinamit, line 625: I prefer `get` to `check` because this actuallly returns a result

kajinamit (commented):
I'm leaving a few comments but the change is looks good and these can be addressed in follow-up (or you might want to fix the function name before merging it)

kajinamit (approved):
lgtm. Thanks !

controllers/horizon_controller.go
bshephar: Like a memcachedv1.Create() interface or something you mean?

Or maybe we could add something to lib-common if lots of operators will need to implement this.

kajinamit: I'm not too sure about the function name but the same will be needed by the other operators to deploy memcached, so having this logic in a common, reusable space would make sense.

PR #60, opened 2023-03-13
Remove -memcached from memcached resource name

bshephar:
With this patch implemented:
https://github.com/openstack-k8s-operators/infra-operator/pull/14

We no longer need to append the name of the memcached resource with -memcached. This change removes it and just uses the instance.Name instead.


bshephar:
Too much has changed now. I'll submit a new PR for this.

PR #61, opened 2023-03-13
Add API go.mod

bshephar:


api/go.mod
```diff
@@ -0,0 +1,66 @@
+module github.com/openstack-k8s-operators/horizon-operator/api
+
+go 1.20
```
gibizer, line 3: I think the rest of our operators are on golang 1.19. I'm not sure it will cause trouble when integration with openstack-operator.

bshephar: Don't recall setting this. I guess it defaults to whatever my installed go version is. Changed it, and we'll see what conflicts that causes if any.

gibizer (approved):

PR #64, opened 2023-03-13
Implement logic required to render a list of memcached instances for the CACHES config.

bshephar:
This change implements the logic required to render a list of memcached instances. This is then rendered into the local_settings config like so:

```
'LOCATION': [ 'horizon-memcached-0.horizon-memcached:11211','horizon-memcached-1.horizon-memcached:11211','horizon-memcached-2.horizon-memcached:11211' ]
```

This depends-on changes to lib-common here:
https://github.com/openstack-k8s-operators/lib-common/pull/218

Additionally, this fixes a invalid log option.
In the getMemcachedSvc function, we were returning an error if the service couldn't be found. But this required the use of fmt.Sprintf to ensure the error is added to the string. This change adds the missing fmt.Sprintf() to the info log.


controllers/horizon_controller.go
```diff
@@ -502,37 +502,26 @@ func (r *HorizonReconciler) generateServiceConfigMaps(
...
+	memcachedAddress := formatMemcachedAddress(podList)
+
 	url := strings.TrimPrefix(instance.Status.HorizonEndpoints["public"], "http://")
 
 	templateParameters := map[string]interface{}{
 		"keystoneURL":        keystonePublicURL,
 		"horizonDebug":       instance.Spec.Debug,
 		"horizonEndpointUrl": url,
-		"memcachedSvc":       memcachedSvc,
+		"memcachedAddress":   memcachedAddress,
```
kajinamit, line 524: This value contains not only address but also port. Probably memcachedServers would be better

controllers/horizon_controller.go
```diff
@@ -625,22 +614,30 @@ func (r *HorizonReconciler) renderMemcached(instance *horizonv1alpha1.Horizon) *
...
+	if err != nil {
+		return nil, err
 	}
-	return "", fmt.Errorf("No memcached service was found")
+	return podList, nil
 }
 
-func validateHorizonSecret(secret *corev1.Secret) bool {
-	return len(secret.Data["horizon-secret"]) != 0
+func formatMemcachedAddress(podList []string) string {
```
kajinamit, line 636: formatMemcachedServers(hostList []string)

controllers/horizon_controller.go
```diff
@@ -625,22 +614,30 @@ func (r *HorizonReconciler) renderMemcached(instance *horizonv1alpha1.Horizon) *
...
-	labelSelector := k8s_labels.Set{"app": "memcached"}
-	svcList, err := oko_svc.GetServicesListWithLabel(ctx, h, namespace, labelSelector)
+func validateHorizonSecret(secret *corev1.Secret) bool {
+	return len(secret.Data["horizon-secret"]) != 0
+}
 
-	if err != nil {
-		r.Log.Info("Error getting service list: %s", err)
-		return "", err
+func getMemcachedPods(ctx context.Context, h *helper.Helper, instance *horizonv1alpha1.Horizon, memcachedName string) ([]string, error) {
```
kajinamit, line 621: getMemcachedHosts would be better hear because it returns list of hostnames

controllers/horizon_controller.go
```diff
@@ -625,22 +614,30 @@ func (r *HorizonReconciler) renderMemcached(instance *horizonv1alpha1.Horizon) *
...
+	labelSelector := map[string]string{
+		"app":            "memcached",
+		"memcached/name": memcachedName,
 	}
-	for _, service := range svcList.Items {
-		if service.Name == memcachedName {
-			return service.Name, nil
-		}
+
+	podList, err := oko_pod.GetPodFQDNList(ctx, h, instance.Namespace, labelSelector)
```
kajinamit, line 629: hostList

controllers/horizon_controller.go
```diff
@@ -474,7 +474,7 @@ func (r *HorizonReconciler) generateServiceConfigMaps(
 	h *helper.Helper,
 	envVars *map[string]env.Setter,
 ) error {
-	var memcachedSvc string
+	var podList []string
```
kajinamit, line 477: memcachedHostList

controllers/horizon_controller.go
```diff
@@ -625,22 +614,30 @@ func (r *HorizonReconciler) renderMemcached(instance *horizonv1alpha1.Horizon) *
 	}
 }
 
-func (r *HorizonReconciler) getMemcachedSvc(ctx context.Context, memcachedName string, h *helper.Helper, namespace string) (svc string, err error) {
-	labelSelector := k8s_labels.Set{"app": "memcached"}
-	svcList, err := oko_svc.GetServicesListWithLabel(ctx, h, namespace, labelSelector)
+func validateHorizonSecret(secret *corev1.Secret) bool {
```
kajinamit, line 617: seems this is unused and can be a separate PR ?

kajinamit:
One thing I noticed. Should we add the rbac rule to allow horizon controller to get pod infromation ? That's still missing iiuc.

controllers/horizon_controller.go
bshephar: I feel conflicted about this. It makes sense from the memcached perspective, but in terms of Kubernetes using HostList seems to imply I'm getting information about the actual OpenShift nodes rather than a pod. Maybe I could use a longer variable name to clarify exactly what this is? Something like `podFqdnList` or something? `podHostnameList`?

controllers/horizon_controller.go
bshephar: ack. Sounds better.

controllers/horizon_controller.go
bshephar: `getMemcachedPodFQDNs`?

kajinamit (approved):
lgtm

PR #71, opened 2023-03-22
Re-create secret if for any reason it's empty

bshephar:
If we ever find that the Horizon secret is empty, this change will ensure that the secret is recreated during the reconcile loop. This avoids situations where Horizon will fail to start due to the secret being empty.

### DIFF
--- a/controllers/horizon_controller.go
+++ b/controllers/horizon_controller.go
@@ -586,10 +586,10 @@ func (r *HorizonReconciler) ensureHorizonSecret(
 	//
 	// check if secret already exist
 	//
-	_, _, err := oko_secret.GetSecret(ctx, h, horizon.ServiceName, instance.Namespace)
+	scrt, _, err := oko_secret.GetSecret(ctx, h, horizon.ServiceName, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return err
-	} else if k8s_errors.IsNotFound(err) {
+	} else if k8s_errors.IsNotFound(err) || !validateHorizonSecret(scrt) {
 		r.Log.Info("Creating Horizon Secret")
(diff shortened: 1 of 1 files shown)


controllers/horizon_controller.go
```diff
@@ -586,10 +586,10 @@ func (r *HorizonReconciler) ensureHorizonSecret(
 	//
 	// check if secret already exist
 	//
-	_, _, err := oko_secret.GetSecret(ctx, h, horizon.ServiceName, instance.Namespace)
+	scrt, _, err := oko_secret.GetSecret(ctx, h, horizon.ServiceName, instance.Namespace)
 	if err != nil && !k8s_errors.IsNotFound(err) {
 		return err
-	} else if k8s_errors.IsNotFound(err) {
+	} else if k8s_errors.IsNotFound(err) || len(scrt.Data["horizon-secret"]) == 0 {
```
kajinamit, line 592: Should we fail in case secret is empty ? I expect some users might want to inject their own secret manually then we should warn the user that the given secret is incorrect instead of overriding it silently.

kajinamit: Also, it might be helpful for future maintenance if we can create a separate validateHorizonSecret function to gather all logics we need for validation.

bshephar: So, I updated just the function part at the moment. Moved the check into a function like you suggested.

To ensure the conversation is also captured here for posterity. We will implement a mechanism that will allow a user to manually define this secret via the Spec. In which case, we will not override user provided secrets. Such a check will need to validate that:

1. The secret exists;
2. The secret IS owned by the operator

Else, we will fail and update the `Status.Condition` to make users aware of invalid Horizon secrets, such as empty ones in this case.

kajinamit (approved):

PR #78, opened 2023-03-30
Remove duplicate vhost configs

kajinamit:
... and use the base single file to define all apache configurations.


bshephar (commented):
Ah yes, my mistake. This looks good, but for some reason it's updating the CRD templates in pre-commit. Maybe `make manifests` is updating something during pre-commit.

bshephar (approved):

PR #79, opened 2023-03-31
Enable readiness probe

kajinamit:
... so that we can ensure the service is ready before we start sending traffic to a pod.

### DIFF
--- a/pkg/horizon/deployment.go
+++ b/pkg/horizon/deployment.go
@@ -48,6 +48,17 @@ func Deployment(instance *horizonv1.Horizon, configHash string, labels map[strin
 			},
 		},
 	}
+	readinessProbe := &corev1.Probe{
+		TimeoutSeconds:      5,
+		PeriodSeconds:       10,
+		InitialDelaySeconds: 10,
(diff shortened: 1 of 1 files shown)


pkg/horizon/deployment.go
```diff
@@ -85,10 +96,11 @@ func Deployment(instance *horizonv1.Horizon, configHash string, labels map[strin
...
 								RunAsUser: &runAsUser,
 							},
-							Env:           env.MergeEnvs([]corev1.EnvVar{}, envVars),
-							VolumeMounts:  getVolumeMounts(),
-							Resources:     instance.Spec.Resources,
-							LivenessProbe: livenessProbe,
+							Env:            env.MergeEnvs([]corev1.EnvVar{}, envVars),
+							VolumeMounts:   getVolumeMounts(),
+							Resources:      instance.Spec.Resources,
+							ReadinessProbe: ReadinessProbe,
```
bshephar, line 102: upper case vs lower case in the variable declaration I believe is the only issue here.

bshephar (approved):

PR #82, opened 2023-04-06
Remove the -memcached prefix

bshephar:
When we create a instance of memcached, we no longer need to append it with -memcached. This change removes that and just uses the instance name instead.


abays (approved):
/lgtm

PR #101, opened 2023-04-20
Add revive to golanci

bshewale:
This PR replace deprecated go-linter using
revive (https://golangci-lint.run/usage/linters/#revive).


kajinamit:
OK I think https://github.com/openstack-k8s-operators/horizon-operator/pull/108 should resolve the lint error.

PR #110, opened 2023-05-04
Webhooks for default images

abays:
Default container images are now handled via env vars and internal webhooks, as per the pattern we are using for all NextGen operators.  This PR also changes the Horizon `api` version to `v1beta1` to match our other operators.


bshephar:
I'm not a huge fan of the abrupt change to v1beta1 and removal of v1alpha1. Feels like something like that should have been its own PR, but since we're still deep in the development phase of this project, I guess it should be ok to include here.

We will need to also update openstack-operator if we're going to get rid of v1alpha1 since it currently imports it. So if we merge this, I think openstack-operator will complain until we fix it there as well.

kajinamit:
/lgtm

Let's merge this and work on migration to v1beta, because we aim to bump version before GA.

PR #113, opened 2023-05-10
Add kuttl tests

jparoly:


tests/kuttl/tests/01-assert-deploy-horizon.yaml
abays, line 1: There needs to be an encapsulating directory for this file and the `01-deploy-horizon.yaml` file, which would signify the name of the test.  Otherwise KUTTL looks at the `tests/kuttl/tests` dir and see no sub-dirs and assumes there are no tests to run.

tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: horizon.openstack.org/v1beta1
```
abays, line 1: `v1beta1` needs to be `v1alpha1`, as this is the current version of the `Horizon` API

kuttl-test.yaml
```diff
@@ -0,0 +1,23 @@
...
+#    - chmod 755 /usr/local/bin/kubectl-kuttl
+# 2. An OCP 4.10+ CRC cluster with Podified Operators has been deployed
+# 3. CLI user has access to $KUBECONFIG
+
+apiVersion: kuttl.dev/v1beta1
+kind: TestSuite
+reportFormat: JSON
+reportName: kuttl-test-horizon
+timeout: 180
+namespace: horizon-kuttl-tests
```
abays, line 20: For the moment, let's switch this namespace to `openstack`, as that is what is used elsewhere by the tests (otherwise the current test will fail).  We can follow-up later when we sweep through operators to support arbitrary-namespace KUTTL tests.

kuttl-test.yaml
```diff
@@ -0,0 +1,23 @@
+#
+# EXECUTION (from install_yamls repo root):
+#
+#   make keystone_kuttl
```
abays, line 4: `s/keystone/horizon`

abays (approved):
/lgtm

abays (approved):
/lgtm

PR #114, opened 2023-05-11
Fix remaining references to old API version

kajinamit:
We have bumped API version of this operator from v1alpha to v1beta. This fixes a few remaining references in files for kuttl tests.


abays (approved):
/lgtm

PR #115, opened 2023-05-11
Add container image to CR sample

cescgina:


cescgina:
Closing this as it's not really needed since the container image is provided via the webhook.

PR #134, opened 2023-06-05
Display status in CLI

kajinamit:
Allows `Ready` condition status to be display at the CLI via `oc get horizon`, similar to what we do for other operators.

This follows what was done for heat-operator by https://github.com/openstack-k8s-operators/heat-operator/pull/138 .


abays (approved):
/lgtm

PR #136, opened 2023-06-05
Assert Memcached CR in kuttl tests

kajinamit:
Horizon controller creates Memcached for caching. This makes sure the CR instance is created.


abays (approved):
/lgtm

PR #137, opened 2023-06-05
Use constant to look up endpoints

kajinamit:
... instead of hard-coding the string.

### DIFF
--- a/controllers/horizon_controller.go
+++ b/controllers/horizon_controller.go
@@ -266,7 +266,7 @@ func (r *HorizonReconciler) reconcileInit(
 	//
 	// Update instance status with service endpoint url from route host information
 	//
-	instance.Status.Endpoint = apiEndpoints["public"]
+	instance.Status.Endpoint = apiEndpoints[string(endpoint.EndpointPublic)]
 
 	// expose service - end
 


controllers/horizon_controller.go
```diff
@@ -266,7 +266,7 @@ func (r *HorizonReconciler) reconcileInit(
 	//
 	// Update instance status with service endpoint url from route host information
 	//
-	instance.Status.Endpoint = apiEndpoints["public"]
+	instance.Status.Endpoint = apiEndpoints[endpoint.PublicEndpoint]
```
kajinamit, line 269: Alternatively we can cast this to string but that is still redundant.

```
instance.Status.Endpoint = apiEndpoints[string(endpoint.PublicEndpoint)]
```

bshephar, line 269: `endpoint.EndpointPublic`
https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/endpoint/endpoint.go#L42

bshephar (changes requested):

controllers/horizon_controller.go
kajinamit: :facepalm: Fixed it. Also probably we can cast this to string explicitly for now until the lib-common change is approved.

bshephar: ack. Do you have the link to a lib-common PR for that change? Or just being discussed atm?

bshephar (approved):

PR #138, opened 2023-06-07
Add memcached suffix for Memcached CR name

kajinamit:
Some resources created for Memcached CR do not contain prefix/suffix and use instance name. So adding the suffix, following how we create transport url, would be needed to help users identify resources created for Memcached CR.


abays (approved):
/lgtm

PR #145, opened 2023-06-22
Stop deploying memcached automatically

kajinamit:
... but require an pre-deployed instance. This follows the current design to accept mariadb database or rabbitmq.

Depends-on: https://github.com/openstack-k8s-operators/install_yamls/pull/338

### DIFF
--- a/controllers/horizon_controller.go
+++ b/controllers/horizon_controller.go
@@ -44,7 +44,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/rand"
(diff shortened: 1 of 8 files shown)


config/samples/horizon_v1beta1_horizon.yaml
```diff
@@ -7,3 +7,11 @@ spec:
   secret: "osp-secret"
   customServiceConfig: |
     SESSION_TIMEOUT = 3600
+  memcachedInstance: horizon-memcached
+---
+apiVersion: memcached.openstack.org/v1beta1
```
kajinamit, line 12: this should be later removed when install_yams deploy a separate memcached instance.

controllers/horizon_controller.go
```diff
@@ -347,21 +346,15 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
...
 
 	// Create Memcached instance if no shared instance exists.
-	var memcached *memcachedv1.Memcached
-	if instance.Spec.SharedMemcached == "" {
-		memcached, err = r.ensureHorizonMemcached(ctx, helper, instance)
-	} else {
-		memcached, err = r.getSharedMemcached(ctx, helper, instance)
-	}
-
+	memcached, err := r.getHorizonMemcached(ctx, helper, instance)
```
kajinamit, line 349: We can probably move this into generateHorizonConfigMap and treat memcached similarly to KeystoneAPI ?

config/samples/horizon_v1beta1_horizon.yaml
kajinamit: We probably should merge https://github.com/openstack-k8s-operators/install_yamls/pull/338 first.

controllers/horizon_controller.go
bshephar: Yeah, I think we can move it now that it's been simplified. I think I put it in `reconcileNormal()` initially because it needed to create the Memcached instance before anything else could succeed.

bshephar (commented):
Not sure why kuttl is failing here. It seems to be complaining that there is no readyCount key:
```
case.go:366: resource Horizon:openstack/horizon: .status.readyCount: key is missing from map
```

kajinamit:
> 

I think this is because memached is not deployed. We have to merge https://github.com/openstack-k8s-operators/install_yamls/pull/338 first.

controllers/horizon_controller.go
kajinamit: A problem I noticed while moving this logic is that `generateServiceConfigMaps` now supports only immediate failure and to support retry after some period we have to modify its interface and also add a new status for ServiceConfigReadyCondition... I probably pend that refactoring and distribute this implementation as it is now.

bshephar (approved):
Looks good to me. I noticed some of the other operators are starting to use `namev1` instead of `namev1beta1` now. We can probably look at changing that for Horizon too. 

But this looks good. A nice addition, and a good culmination to all of the work you have put in to standardising the way we interact with memcached resources.

PR #148, opened 2023-06-25
Add ability to set Horizon route annotation

bshephar:
This change adds the ability to set annotations on the Horizon route. This can be leveraged as per the OpenShift documentation: https://docs.openshift.com/container-platform/4.13/networking/routes/route-configuration.html#nw-route-specific-annotations_route-configuration

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/288


tests/functional/horizon_controller_test.go
```diff
@@ -145,4 +146,21 @@ var _ = Describe("Horizon controller", func() {
...
+					Namespace: namespace,
+				},
+			}
+			Expect(k8sClient.Create(ctx, secret)).Should(Succeed())
+		})
+		It("Should have an Annotation on the Route", func() {
+			horizonRoute := k8sClient.List(ctx, &routev1.RouteList{})
+			fmt.Printf("debug: HorizonRoute: %+v", horizonRoute)
+		})
+	})
```
bshephar, line 165: I don't think this will be possible with Envtest. The Route API is provided via the `openshift-api` server using the API Aggregation Kubernetes method:
https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/apiserver-aggregation/

So, unless we start Envtest and install the Openshift-api server, I don't know that we will have access to any of the custom resources provided by Openshift-api.

bshephar: Again, this is where it would be easier to just use Ingress...

bshephar, line 165: Need to use the Route helper from lib-common to test this:
https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/test/openshift_crds/README.md

PR #149, opened 2023-06-26
Fix inconsistent package name for functional tests

kajinamit:
The package name for functional tests is functional_test instead of functional in most of our operators. Let's use the consistent name.


abays (approved):
/lgtm

PR #150, opened 2023-06-28
Move RBAC config into it's own function

bshephar:
Largely, reconcileNormal() is used to call functions and set status conditions. This change moves the logic for rbac into it's own function and returns the result or error which is more consistent with the rest of the reconcileNormal() tasks.

Additionally, fmt.Errorf strings should be lower case: https://google.github.io/styleguide/go/decisions.html#error-strings


abays (approved):
/lgtm

PR #152, opened 2023-06-29
Remove handling for osp-secret

bshephar:
Horizon uses it's own secret to store the SECRET_KEY. This is done outside of the general purpose OSP Secret. This change removes handling for the general osp-secret.

### DIFF
--- a/api/v1beta1/horizon_types.go
+++ b/api/v1beta1/horizon_types.go
@@ -44,10 +44,6 @@ type HorizonSpec struct {
 	// Replicas of horizon API to run
 	Replicas int32 `json:"replicas"`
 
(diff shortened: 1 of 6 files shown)


kajinamit:
I'm ok with the change but tend to keep it for a while until we have better view about the overall feature we need.
For example we likely need to add this back in case we want to support mysql session storage(instead of memcached).

bshephar:
> I'm ok with the change but tend to keep it for a while until we have better view about the overall feature we need. For example we likely need to add this back in case we want to support mysql session storage(instead of memcached).

Maybe we can ask the Horizon folks about their intentions there.

PR #154, opened 2023-06-30
Fix a few missing delete rbac rules

kajinamit:
horizon operator should be able to delete the resources it creates.


abays (approved):
/lgtm

PR #156, opened 2023-07-03
Extend liveness probe timing

bshephar:
The Horizon pod is fairly frequently killed due to the current liveness probe timing. This seems to happen more frequently during the initial deployment while the system is under load. This change extends the liveness probe to provide some extra time for kubelet to validate the functionality of the Horizon pod before terminating it.

### DIFF
--- a/pkg/horizon/deployment.go
+++ b/pkg/horizon/deployment.go
@@ -38,9 +38,9 @@ func Deployment(instance *horizonv1.Horizon, configHash string, labels map[strin
 
 	args := []string{"-c", ServiceCommand}
 	livenessProbe := &corev1.Probe{
-		TimeoutSeconds:      5,
-		PeriodSeconds:       10,
-		InitialDelaySeconds: 10,
+		TimeoutSeconds:      15,
+		PeriodSeconds:       5,
+		InitialDelaySeconds: 15,
 		ProbeHandler: corev1.ProbeHandler{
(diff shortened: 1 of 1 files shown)


abays:
> The Horizon pod is fairly frequently killed due to the current liveness probe timing. This seems to happen more frequently during the initial deployment while the system is under load. This change extends the liveness probe to provide some extra time for kubelet to validate the functionality of the Horizon pod before terminating it.

If this failure happens during the initial deployment, I wonder if it would be prudent to consider adding a startup probe [1]?

[1] https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

bshephar:
> > The Horizon pod is fairly frequently killed due to the current liveness probe timing. This seems to happen more frequently during the initial deployment while the system is under load. This change extends the liveness probe to provide some extra time for kubelet to validate the functionality of the Horizon pod before terminating it.
> 
> If this failure happens during the initial deployment, I wonder if it would be prudent to consider adding a startup probe [1]?
> 
> [1] https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

Sounds reasonable. Let's give it a run and see how it goes. I changed my commit to reflect this.

abays (approved):
/lgtm

PR #158, opened 2023-07-04
Use DeleteInstance from lib-common test helper

kajinamit:
... instead of maintaining duplicated implementations in this repository.


bshephar (approved):
Nice one. Good to reduce the code that we need to maintain independently.

PR #159, opened 2023-07-04
Allow overriding Horizon Spec

bshephar:
This refactors the functions for creating the Heat instance to allow for the possibility of overriding the default horizon.HorizonSpec

### DIFF
--- a/tests/functional/base_test.go
+++ b/tests/functional/base_test.go
@@ -30,8 +30,8 @@ import (
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
-func DefaultHorizonTemplate(name types.NamespacedName) *horizon.Horizon {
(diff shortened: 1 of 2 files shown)


tests/functional/base_test.go
```diff
@@ -30,7 +30,7 @@ import (
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 )
 
-func DefaultHorizonTemplate(name types.NamespacedName) *horizon.Horizon {
+func DefaultHorizonTemplate(name types.NamespacedName, spec horizon.HorizonSpec) *horizon.Horizon {
```
kajinamit, line 33: wondering if we can merge this into CreateHorizon following the interface we have in heat-operator.

bshephar: done

PR #164, opened 2023-07-06
Add kuttl test assert to check Horizon route

bshephar:
This change adds a Kuttl test assertion to ensure the Horizon route is responding to curl requests with the appropriate HTTP status codes.

### DIFF
--- a/tests/kuttl/common/assert-sample-deployment.yaml
+++ b/tests/kuttl/common/assert-sample-deployment.yaml
@@ -9,3 +9,19 @@ spec:
     SESSION_TIMEOUT = 3600
 status:
   readyCount: 1
+---
+# Test the status code is correct for each endpoint
+# This test is for heat endpoints
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+      set -x
+      RETURN_CODE=0
+      PUBLIC_URL=$(oc get -n $NAMESPACE horizon horizon -o jsonpath='{.status.endpoint}')
+      STATUSCODE=$(curl --silent --output /dev/stderr --head --write-out "%{http_code}" "$PUBLIC_URL/dashboard/auth/login/?next=/dashboard/")
+      if test $STATUSCODE -ne 200; then
+          RETURN_CODE=1
+          echo "${PUBLIC_URL} status code expected is 200 but was ${STATUSCODE}"
+      fi
+      exit $RETURN_CODE


tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -9,3 +9,19 @@ spec:
...
+---
+# Test the status code is correct for each endpoint
+# This test is for heat endpoints
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+      set -x
+      RETURN_CODE=0
+      PUBLIC_URL=$(oc get route -n $NAMESPACE horizon-public -o jsonpath='{.status.ingress[0].host}')
```
kajinamit, line 21: I think we can use Status.Endpoint of Horizon CR instead.

tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -9,3 +9,19 @@ spec:
...
+      STATUSCODE=$(curl --silent --output /dev/stderr --head --write-out "%{http_code}" $PUBLIC_URL/healthcheck)
```
kajinamit, line 22: Do we really have this path available ? This is provided by the healthcheck middleware used by OpenStack services but I don't think we have it for horizon.

bshephar: Ah true. I copied from another operator. I'll fix this.

tests/kuttl/common/assert-sample-deployment.yaml
bshephar: Yeah, we could do that. I just copied this from Heat. But we can do it either way I guess.

bshephar: Ok, updated to use the `status.endpoint`

tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -9,3 +9,19 @@ spec:
...
+# Test the status code is correct for each endpoint
+# This test is for heat endpoints
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+      set -x
+      RETURN_CODE=0
+      PUBLIC_URL=$(oc get -n $NAMESPACE horizon horizon -o jsonpath='{.status.endpoint}')
+      STATUSCODE=$(curl --silent --output /dev/stderr --head --write-out "%{http_code}" $PUBLIC_URL/dashboard/auth/login/?next=/dashboard/")
```
kajinamit, line 22: Looks like you should add `"` before `$PUBLIC_URL` ?

kajinamit (changes requested):
The failure in kuttl test looks valid. See my comment inline.

tests/kuttl/common/assert-sample-deployment.yaml
bshephar: Ah yeah, I see where I missed it. Fixed in the latest revision.

kajinamit:
/test horizon-operator-build-deploy-kuttl

We probably have to rerun it after we merged the change required to run kuttl in a dedicated namespace.

tests/kuttl/common/assert-sample-deployment.yaml
```diff
@@ -9,3 +9,19 @@ spec:
...
+# This test is for heat endpoints
+apiVersion: kuttl.dev/v1beta1
+kind: TestAssert
+commands:
+  - script: |
+      set -x
+      RETURN_CODE=0
+      PUBLIC_URL=$(oc get -n $NAMESPACE horizon horizon -o jsonpath='{.status.endpoint}')
+      STATUSCODE=$(curl --silent --output /dev/stderr --head --write-out "%{http_code}" "$PUBLIC_URL/dashboard/auth/login/?next=/dashboard/")
+      if test $STATUSCODE -ne 204; then
```
kajinamit, line 23: I think the expected status code is 200 instead of 204.

kajinamit (approved):
/lgtm

PR #168, opened 2023-07-06
Remove unnecessary rbac rules about Memcached

kajinamit:
The horizon controller no longer deploys Memcached but only look up the existing instance.


abays (approved):
/lgtm

PR #170, opened 2023-07-06
Watch for Memcached

kajinamit:
Since we now pass a `Memcached` instance name to the `Horizon`, its controller should watch for changes to `Memcacheds`.

This imports https://github.com/openstack-k8s-operators/keystone-operator/pull/271 in keystone-operator.


abays (approved):
/lgtm

PR #173, opened 2023-07-10
Switch to structured logging

kajinamit:
This replaces the current logging by structured logging, following the change earlier made in keystone-operator[1].

[1] https://github.com/openstack-k8s-operators/keystone-operator/pull/220

### DIFF
--- a/controllers/horizon_controller.go
+++ b/controllers/horizon_controller.go
@@ -67,21 +67,20 @@ func (r *HorizonReconciler) GetKClient() kubernetes.Interface {
 	return r.Kclient
 }
 
-// GetLogger -
-func (r *HorizonReconciler) GetLogger() logr.Logger {
-	return r.Log
-}
-
 // GetScheme -
 func (r *HorizonReconciler) GetScheme() *runtime.Scheme {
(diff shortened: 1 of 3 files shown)


controllers/horizon_controller.go
```diff
@@ -251,11 +249,12 @@ func (r *HorizonReconciler) SetupWithManager(mgr ctrl.Manager) error {
 }
 
 func (r *HorizonReconciler) reconcileDelete(ctx context.Context, instance *horizonv1beta1.Horizon, helper *helper.Helper) (ctrl.Result, error) {
-	r.Log.Info("Reconciling Service delete")
+	l := GetLog(ctx)
```
bshephar, line 252: Might be nice to keep a consistent name for this. Either `l` or `logger`. Makes it easier to find log messages with regex searching at least if the names are always consistent.

kajinamit: This follows the existing naming in keystone-operator, and I think it makes sense because `l` and `logger` are initialized differently.

bshephar (approved):

PR #174, opened 2023-07-10
Use arbitrary namespace for kuttl tests

kajinamit:
... instead of using the openstack namespace which is now globally used.

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -16,8 +16,8 @@ apiVersion: kuttl.dev/v1alpha1
 kind: TestSuite
 reportFormat: JSON
 reportName: kuttl-test-horizon
+namespace: horizon-kuttl-tests
 timeout: 180
-namespace: openstack
 parallel: 1
 suppress:
   - events                     # Remove spammy event logs
--- a/tests/kuttl/tests/basic-deployment/01-deploy-horizon.yaml
+++ b/tests/kuttl/tests/basic-deployment/01-deploy-horizon.yaml
@@ -2,4 +2,4 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestStep
 commands:
   - script: |
-      oc apply -n openstack -f ../../../../config/samples/horizon_v1beta1_horizon.yaml
+      oc apply -n $NAMESPACE -f ../../../../config/samples/horizon_v1beta1_horizon.yaml


kuttl-test.yaml
```diff
@@ -17,7 +17,6 @@ kind: TestSuite
 reportFormat: JSON
 reportName: kuttl-test-horizon
 timeout: 180
-namespace: openstack
```
abays, line 20: I think the pattern we have been using is like so [1] instead of removing the field completely.

[1] https://github.com/openstack-k8s-operators/keystone-operator/blob/fab7648d6c1074e9f82b6b21ba4b2a5310163f2e/kuttl-test.yaml#L20

kajinamit: Hmm. We removed this in heat-operator instead of updating it. https://github.com/openstack-k8s-operators/heat-operator/commit/02940fab5059178a80e980963e42bad7920cd508 

Let me update this PR and heat-operator.

Note that, as far as I've tested, it seems NAMESPACE environment overrides the setting in `kuttl-test.yaml` so probably we should merge the change of install_yamls first.

abays: Okay, I will look at the `install-yamls` PR

kajinamit: I had conversation about this with @cescgina , and I learned this namespace is not really used within jobs as we pass --namespace option as I'm adding here https://github.com/openstack-k8s-operators/install_yamls/pull/388/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R1406 .

We agreed that kuttl namespace as the default one we want to use, so the current PR should be correct.

cescgina:
The namespace looks correct, it looks like we still need to replace a mention to `openstack` by the `$NAMESPACE` variable in https://github.com/kajinamit/horizon-operator/blob/8445a37ad04fc4287f53e86e0c40adbbf56d509d/tests/kuttl/tests/basic-deployment/01-deploy-horizon.yaml#L5

kajinamit:
> The namespace looks correct, it looks like we still need to replace a mention to `openstack` by the `$NAMESPACE` variable in https://github.com/kajinamit/horizon-operator/blob/8445a37ad04fc4287f53e86e0c40adbbf56d509d/tests/kuttl/tests/basic-deployment/01-deploy-horizon.yaml#L5

thanks. Fixed that.

kajinamit:
So I tested this with https://github.com/openstack-k8s-operators/install_yamls/pull/388 locally and confirmed kuttl job passes. I guess we have to merge that install_yamls change first.

cescgina:
> So I tested this with [openstack-k8s-operators/install_yamls#388](https://github.com/openstack-k8s-operators/install_yamls/pull/388) locally and confirmed kuttl job passes. We have to merge that install_yamls change first in case both fails at CI (because of a kind of circular dependency).

Yes, the install_yamls change needs to go in first, so the dependencies are installed in the `horizon-kuttl-tests` namespaces. After that the kuttl prow job here should pass

bshephar (approved):
Based on the discussion that has been resolved here, let's go ahead with it.

PR #178, opened 2023-07-12
Remove duplicate jobs from github workflow

kajinamit:
We now run pre-commit job in prow which covers most of go-lang tasks.


.github/workflows/lints.yaml
```diff
@@ -0,0 +1,14 @@
...
+jobs:
+  check-go-mod-replace-lines:
+    name: check for replace lines in go.mod files
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout project code
+        uses: actions/checkout@v3
+      - name: check for replace lines in go.mod files
+        run: |
+          ! egrep --invert-match -e '^replace.*/api => \./api|^replace.*//allow-merging$'  `find . -name 'go.mod'` | egrep -e 'go.mod:replace'
```
bshephar, line 14: Just noting my comment here as well to carry over our private conversation.

We might need to keep this in mind if we ever encounter a situation where we need to pin dependency versions to a certain release. Which does come up in lots of other Kubernetes projects. We definitely don't have that situation right now, so we can go ahead with this. But as the project matures, we may need to adjust course on this.

bshephar (approved):

PR #179, opened 2023-07-14
Ensure webhook is started

kajinamit:
This ensures webhook is started when controller-manage deployment becomes ready.

### DIFF
--- a/main.go
+++ b/main.go
@@ -126,20 +126,22 @@ func main() {
 	// Acquire environmental defaults and initialize operator defaults with them
 	horizonv1beta1.SetupDefaults()
 
+	checker := healthz.Ping
 	// Setup webhooks if requested
 	if strings.ToLower(os.Getenv("ENABLE_WEBHOOKS")) != "false" {
 		if err = (&horizonv1beta1.Horizon{}).SetupWebhookWithManager(mgr); err != nil {
 			setupLog.Error(err, "unable to create webhook", "webhook", "Horizon")
 			os.Exit(1)
 		}
+		checker = mgr.GetWebhookServer().StartedChecker()
 	}
 	//+kubebuilder:scaffold:builder
 
-	if err := mgr.AddHealthzCheck("healthz", healthz.Ping); err != nil {
+	if err := mgr.AddHealthzCheck("healthz", checker); err != nil {
 		setupLog.Error(err, "unable to set up health check")
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err := mgr.AddReadyzCheck("readyz", checker); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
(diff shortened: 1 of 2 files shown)


bshephar (approved):

abays:
Is there any concern here that now we'll assume the manager pod is ready before the actual reconciler is available?  i.e. just examining an example log:

```
2023-07-13T18:11:36Z    INFO    controller-runtime.webhook      Serving webhook server  {"host": "", "port": 9443}
2023-07-13T18:11:36Z    INFO    controller-runtime.certwatcher  Starting certificate watcher
I0713 18:11:55.500806       1 leaderelection.go:258] successfully acquired lease openstack-operators/5ad2eba0.openstack.org
2023-07-13T18:11:55Z    DEBUG   events  horizon-operator-controller-manager-7b57db86b5-qt6dk_636f6e13-4d1a-48a0-99d8-3ab449940e6c became leader {"type": "Normal", "object": {"kind":"Lease","namespace":"openstack-operators","name":"5ad2eba0.openstack.org","uid":"d5d94b79-1b5c-4ef8-aed3-9c48283a604e","apiVersion":"coordination.k8s.io/v1","resourceVersion":"7779317"}, "reason": "LeaderElection"}
...
2023-07-13T18:11:56Z    INFO    Starting workers        {"controller": "horizon", "controllerGroup": "horizon.openstack.org", "controllerKind": "Horizon", "worker count": 1}
```

It seems that the webhook server was ready ~20 seconds before the actual reconciler was.  Or does it not matter that the reconciler isn't ready?  Anyhow, just wondering if this might have any side effects.

kajinamit:
Even without this change, manage pod becomes ready regardless of status of the reconciler. So I think that is a different topic. I think it's not required to wait until the actual reconciler becomes available, as long as api can accepts request and reconciler starts later and processes requests.

Without this change, operator deployment becomes ready before webhook is ready, and that makes csv install status succeeeded even before webhook is ready. Because of this behavior we have sometimes seen the api failures like the example below in CI.
```
Internal error occurred: failed calling webhook "mnetconfig.kb.io": failed to call webhook: Post "https://infra-operator-controller-manager-service.openstack-operators.svc:443/mutate-network-openstack-org-v1beta1-netconfig?timeout=10s": dial tcp 10.217.0.93:9443: connect: connection refused
```

abays:
Do we know exactly what the old health/ready check was looking at, then?  If it's not considering the actual reconciler, what is "ready" as far as the default-scaffolded check was concerned?

kajinamit:
If you look at spec of controller-manager deployment, it's checking /healthz and /readyz path to determine readiness/liveness of the pod/deployment.
https://github.com/openstack-k8s-operators/horizon-operator/blob/main/config/manager/manager.yaml#L83-L94

Before this change, we were using `healthz.Ping`, which returns always True. So pod becomes ready/live once the logic to set up these checks are reached.

https://github.com/openstack-k8s-operators/horizon-operator/blob/cfceea907bdd4daf5856d05877d053431e63129a/main.go#L138-L145

After this change, now these checks are backed by the webhook server checker, which tries to connect the webhook server using tcp. So controller-manager pod becomes ready/live only when webhook server is ready and connectable.

PR #193, opened 2023-07-27
Ensure horizon debug log is enabled

kajinamit:
Currently debug log of horizon is not enabled because the log level is determined before DEBUG is overridden.

This adds an additional custom setting so that log level is overridden properly.

### DIFF
--- a/config/samples/horizon_v1beta1_horizon.yaml
+++ b/config/samples/horizon_v1beta1_horizon.yaml
@@ -7,3 +7,4 @@ spec:
   secret: "osp-secret"
   customServiceConfig: |
     DEBUG = True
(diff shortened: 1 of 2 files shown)


abays:
This file also needs to be updated: https://github.com/openstack-k8s-operators/horizon-operator/blob/main/tests/kuttl/common/assert-sample-deployment.yaml

kajinamit:
> This file also needs to be updated: https://github.com/openstack-k8s-operators/horizon-operator/blob/main/tests/kuttl/common/assert-sample-deployment.yaml

Thanks. Updated the assertion file as well.

bshephar (approved):

PR #195, opened 2023-07-31
Do not add finalizer if the CR is marked for deletion

kajinamit:
... and also replace the explicit update call by the patching by the defer function.


abays (approved):
/lgtm

PR #196, opened 2023-08-01
Bump k8s version in envtest

kajinamit:
... to the one consistent with the version in dependencies.

This omits the bug fix version so that the latest one of the 1.26 release is selected.


stuggi (approved):
/lgtm

PR #200, opened 2023-08-06
Use consistent method for setting time.Second * n

bshephar:
Currently, we still have a mix of time.Duration and time.Second. This change replaces the remaining time.Duration instances with time.Second * n. For consistency with other projects in the organization.


abays (approved):
Yes, I think this is the proper format, given the `Second` definition is already defined as a `time.Duration`:

```
const time.Second time.Duration = 1000 * Millisecond // 1s
```

/lgtm

PR #201, opened 2023-08-08
Stop registering schemas for default objects

kajinamit:
Currently we register schemas for corev1 and appsv1 but this is incomplete and we need a few more schemas like batchv1. However we have not noticed it because the resources available in k8s by default don't need explicit schema registration.

Let's remove these instead of keeping the incomplete list.


kajinamit:
/retest

failed to wait for the created cluster claim to become ready: timed out waiting for the condition

abays (approved):
/lgtm

PR #204, opened 2023-08-23
Remove config file for mod_ssl

kajinamit:
We are not using this module really in httpd. Encryption should be implemented by more k8s native way instead of mod_ssl in individual apache process.


kajinamit:
We might need to revisit this when we implement internal TLS. However we should avoid maintaining different config specific to Horizon without real requirements (and I'm not aware of any) in that case so I think it's better to remove this uncommon file now.

bshephar (approved):

PR #223, opened 2023-09-29
Add route lookup and update status.endpoint

bshephar:
This commit adds a route lookup and then uses the returned value to update the instance.Status.Endpoint value.

```
❯ oc get horizon horizon -o yaml | yq .status.endpoint
horizon-openstack-operators.apps.okd.bne-shift.net
```

Before this change, it is defaulting to the Kubernetes service name which doesn't provide much value for the end user.


stuggi:
we don't want the service operators to depend on routes if possible. https://github.com/openstack-k8s-operators/horizon-operator/pull/224 was missing in horizon operator that the openstack-operator correct passes in the endpointURL via the override. if there is no openstack-operator, or if a loadbalancer service gets used for the public endpoints it is correct that the endpoint is a service url

PR #250, opened 2023-11-21
Horizon should use internalURL endpoints

bshephar:
Horizon currently uses the Public endpoints. This causes issues with the current implementation of self-signed cert-manager Issuer. We ultimately need to resolve this trust issue, but in lieu of an established standard for trusting these certs, we will work around this for now by not relying on the public endpoints.


gibizer (approved):

stuggi (approved):
/lgtm

fmount (approved):
/lgtm

PR #260, opened 2023-12-13
Standardize makefile

yazug:


dprince (changes requested):
This one looks different than the others. I think we need to sync it to the latest operator-sdk

yazug:
updated to more recent version of Makefile from operator-sdk skeleton

PR #275, opened 2024-01-30
This change adds requirements for FIPS

bshephar:
This change ensures the heat-operator is built with FIPS compliance.

### DIFF
--- a/.prow_ci.env
+++ b/.prow_ci.env
@@ -1,1 +1,2 @@
 export USE_IMAGE_DIGESTS=true
+export FAIL_FIPS_CHECK=true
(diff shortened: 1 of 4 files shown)


fmount:
The patch itself looks good, we need to add the missing annotation in the clusterservice manifest (see https://github.com/openstack-k8s-operators/docs/blob/main/fips.md#feature-reporting) and then it should be good to go.

fmount (changes requested):
Add:

```
metadata:
  annotations:
    features.operators.openshift.io/fips-compliant: "true"
```
in the clusterserviceversion manifest

fmount (approved):
/lgtm

PR #277, opened 2024-02-02
Clear horizon cached data when config changes

olliewalsh:
The keystone endpoint is cached and may not be valid across config changes. Toggling keystone TLS in particular will change the keystone url and break any existing horizon session.

Jira: [OSPRH-4392](https://issues.redhat.com//browse/OSPRH-4392)


olliewalsh:
doesn't seem to work when using the memcached backend...

olliewalsh:
`django-admin clearsessions` doesn't resolve this either, clearing cookies seems to be the only thing that works

PR #280, opened 2024-02-02
Tweak probe timing

olliewalsh:
Horizon pods are taking a very long time to go to ready state (when the service will start routing traffic to it).

I can see there are init tasks that run on every container startup but they are not that slow.

The StartupProbe initial delay (60s) seem extremely long. Use a 10s delay instead then poll every 10 seconds for up to 5 mins.

Liveness/readiness probes do not run until StartupProbe succeeds so drop the initial delays to 10s.
Also ensure timeout is less that period to prevent hitting the endpoint with multiple concurrent requests.

### DIFF
--- a/pkg/horizon/deployment.go
+++ b/pkg/horizon/deployment.go
@@ -45,9 +45,9 @@ func Deployment(instance *horizonv1.Horizon, configHash string, labels map[strin
 	}
 
 	livenessProbe := &corev1.Probe{
-		TimeoutSeconds:      15,
-		PeriodSeconds:       5,
-		InitialDelaySeconds: 15,
+		TimeoutSeconds:      5,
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

pkg/horizon/deployment.go
```diff
@@ -56,9 +56,9 @@ func Deployment(instance *horizonv1.Horizon, configHash string, labels map[strin
 		},
 	}
 	readinessProbe := &corev1.Probe{
-		TimeoutSeconds:      25,
```
olliewalsh, line 59: I'm assuming the timeout higher than the period wasn't intentional e.g to ensure multiple requests hit the service concurrently. @bshephar do you know?

PR #283, opened 2024-02-14
Controller runtime v0.16.4 Go 1.20

bshephar:


api/go.mod
```diff
@@ -1,70 +1,73 @@
 module github.com/openstack-k8s-operators/horizon-operator/api
 
-go 1.19
+go 1.21
```
bshephar, line 3: Damn, local version is 1.21. I'll need to run go mod tidy with `go1.20.3 mod tidy` I guess

PR #290, opened 2024-02-16
[renovate]Remove golang constraint

gibizer:
The shared global config already has a golang 1.20 constraint


bshephar:
Right, that explains the renovate failures about `go mod tidy`. :)

bshephar (approved):

PR #300, opened 2024-03-14
Fix struct field alignments

bshephar:
This change reorganizes the objects in some of the structs to better align liked sized objects in memory. This reduces the overall memory allocation for each struct by reducing the required padding to align each item.


abays (approved):
/lgtm

PR #307, opened 2024-03-25
Restrict ALLOWED_HOSTS

bshephar:
This change removes the  `ALLOWED_HOSTS=*` and restricts it
instead to the service Endpoint of the Horizon service


deshipu (approved):

abays (approved):
/lgtm

PR #311, opened 2024-04-02
Set tls_context for django/pymemcache

olliewalsh:
Pass a default tls context to pymemcache to use the system-wide CA

Related: [OSPRH-5945](https://issues.redhat.com//browse/OSPRH-5945)


xek (approved):
/lgtm

PR #313, opened 2024-04-08
Set the relevant http headers when Horizon is served as https

deshipu:
There is a number of headers we can enable in Django configuration that make the site somewhat safer.


hjensas (approved):

controllers/horizon_controller.go
```diff
@@ -785,6 +785,7 @@ func (r *HorizonReconciler) generateServiceConfigMaps(
 		"ServerName":         fmt.Sprintf("%s.%s.svc", horizon.ServiceName, instance.Namespace),
 		"Port":               horizon.HorizonPort,
 		"TLS":                false,
+		"isHTTPS":            url.Scheme == "https",
```
olliewalsh, line 788: nit: could make it clearer that this refers to the public endpoint

deshipu: Good point. Updated.

bshephar (approved):

PR #314, opened 2024-04-09
Set ObservedGeneration after condition init

fmount:
This patch implements the pattern that watches at the `ObservedGeneration` to influence the `DeploymentReadyCondition`.
The `ObservedGeneration` is bumped in the main controller after the conditions `Init`, and the deployment compare the generation with the `ObservedGeneration` to make sure we look at the last version of the `Deployment` spec.

### DIFF
--- a/controllers/horizon_controller.go
+++ b/controllers/horizon_controller.go
@@ -178,6 +178,7 @@ func (r *HorizonReconciler) Reconcile(ctx context.Context, req ctrl.Request) (re
 	)
 
 	instance.Status.Conditions.Init(&cl)
+	instance.Status.ObservedGeneration = instance.Generation
 
 	// If we're not deleting this and the service object doesn't have our finalizer, add it.
 	if instance.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(instance, helper.GetFinalizer()) || isNewInstance {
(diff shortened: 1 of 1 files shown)


controllers/horizon_controller.go
```diff
@@ -717,7 +718,8 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
 		return ctrlResult, nil
 	}
 	instance.Status.ReadyCount = depl.GetDeployment().Status.ReadyReplicas
-	if instance.Status.ReadyCount > 0 {
+	if instance.Status.ReadyCount > 0 &&
```
gibizer, line 721: we need to fix this to compare with the Spec.replicas instead of > 0 otherwise the horizon can be ready before all the replicas are updated to the new image

fmount: I think this code is there because of the definition of `DeploymentReadyCondition`: I think we should change it to something like: `instance.Status.ReadyCount == *instance.Spec.Replicas`, which is the same thing we have in other operators.

fmount:
This seems ready to go /cc @gibizer @dprince

gibizer (approved):
looks good

dprince (approved):

gibizer (approved):
still looks good

PR #315, opened 2024-04-09
Support IPv6 in local_settings.py get_pod_ip()

hjensas:
Switch to use socket.getaddrinfo, this supports both IPv4 and IPv6.

OSPRH-6238


hjensas:
So, this is wrong - we want the POD IP - not resolving horizon-openstack.apps ...

hjensas:
> So, this is wrong - we want the POD IP - not resolving horizon-openstack.apps ...

Done.

bshephar (approved):
lgtm thanks!

PR #328, opened 2024-05-29
Enable HSTS on OCP route

bshephar:
This change adds the required annotation to the OCP route to enable HSTS headers as per:
https://access.redhat.com/documentation/en-us/openshift_container_platform/4.7/html/networking/configuring-routes\#nw-enabling-hsts_route-configuration


bshephar:
Implemented in:
https://github.com/openstack-k8s-operators/openstack-operator/pull/819

PR #361, opened 2024-08-28
Add network-attachment-definition support

fmount:
It might be required for the horizon Pods to be able to reach external services on a different network (e.g. mount a `NFS`, exposed using the `storage_network` using `extraMounts`) .
A common use case we had in the previous releases was about using horizon in legacy mode for image upload. In that case the image is temporary stored in a `TEMP_DIR`, which usually has a NFS behind.
For this reason this patch introduces the `NetworkAttachment` interface to the horizon `API` and aligns the operator with the rest of the services. 
By doing this we are able to have an additional interface on the desired network via multus. 

Jira: https://issues.redhat.com/browse/OSPRH-9285


controllers/horizon_controller.go
```diff
@@ -280,6 +283,35 @@ func (r *HorizonReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		return nil
 	}
 
+	// Watch for changes to NADs
+	nadFn := func(ctx context.Context, o client.Object) []reconcile.Request {
+		result := []reconcile.Request{}
+
+		// get all GlanceAPI CRs
+		glanceAPIs := &horizonv1beta1.HorizonList{}
```
bshephar, line 291: Is this copy pasta from Glance and the variable name just hasn't been updated?

```suggestion
		horizons := &horizonv1beta1.HorizonList{}
```

bshephar (changes requested):

controllers/horizon_controller.go
fmount: whoops sorry, bad copy/pasta here :D

fmount: Updated in the last version.

bshephar (approved):
Thanks, this looks good now. Definitely a necessary feature. We'll need some decent test coverage for this feature though.
We can address that at a later time though, since the functional tests added by the PR should be sufficient to ensure we
pass the networkAttachment through to the deployment.

PR #369, opened 2024-09-30
Add kuttl tests to  validate Multus connections

ashu-011:
The horizon-operator is able to accept NetworkAttachment details to connect the Horizon pod to any given Multus network. This is leveraged in cases where Horizon needs access to remote storage. This patch adds kuttl test to validate multus connection is attached properly.


fmount:
IMO this patch can be closed. The current kuttl tests does not deploy an Openstack control plane and they work w/o network isolation. Unless we want to try to add that requirement in install_yamls, I think it's enough to see this feature covered by envTests.
We can add a nad to the openstack-operator sample where net-isolation is defined in case we want to test this scenario.

PR #370, opened 2024-10-04
Check horizon status to validate NetworkAttachment

ashu-011:
This patch modifies the existing functional test to check horizon status to verify multus network exists

Resolves - https://issues.redhat.com/browse/OSPRH-9846

### DIFF
--- a/tests/functional/horizon_controller_test.go
+++ b/tests/functional/horizon_controller_test.go
@@ -413,10 +413,11 @@ var _ = Describe("Horizon controller", func() {
 		})
 	})
 	When("Horizon CR instance is built with NAD", func() {
+		var nad map[string][]string
 		BeforeEach(func() {
 			nad := th.CreateNetworkAttachmentDefinition(types.NamespacedName{
 				Namespace: namespace,
-				Name:      "internalapi",
+				Name:      "storage",
 			})
 			DeferCleanup(th.DeleteInstance, nad)
 			rawSpec := map[string]interface{}{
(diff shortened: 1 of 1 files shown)


tests/functional/horizon_controller_test.go
```diff
@@ -438,7 +438,7 @@ var _ = Describe("Horizon controller", func() {
 		It("Check the resulting endpoints of the generated sub-CRs", func() {
 			th.SimulateDeploymentReplicaReady(horizonName)
 			horizon := GetHorizon(horizonName)
-			Expect(horizon.Spec.NetworkAttachments).To(Equal([]string{"storage"}))
+			Expect(horizon.Status.NetworkAttachments).To(Equal([]string{"storage"}))
```
fmount, line 441: -1: this can't work as you're trying to compare `[]string`  w/ `map[string]string`.
I think the `horizon.Status.NetworkAttachment` assignment is the result of `nad.VerifyNetworkStatusFromAnnotation( ...` in `horizon_controller`[1], which is provided by lib-common [2].
To make it work you need to `th.SimulateDeploymentReadyWithPods(...` and add the annotations that are the result of attaching multus to your `Pods`, and then verify that `horizon.Status.NetworkAttachment` is equals to `nad` (L417).
Here's an example:

```golang
diff --git a/tests/functional/horizon_controller_test.go b/tests/functional/horizon_controller_test.go
index 2ba9275..93237db 100644
--- a/tests/functional/horizon_controller_test.go
+++ b/tests/functional/horizon_controller_test.go
@@ -413,10 +413,11 @@ var _ = Describe("Horizon controller", func() {
                })
        })
        When("Horizon CR instance is built with NAD", func() {
+               var nad map[string][]string
                BeforeEach(func() {
                        nad := th.CreateNetworkAttachmentDefinition(types.NamespacedName{
                                Namespace: namespace,
-                               Name:      "internalapi",
+                               Name:      "storage",
                        })
                        DeferCleanup(th.DeleteInstance, nad)
                        rawSpec := map[string]interface{}{
@@ -434,11 +435,16 @@ var _ = Describe("Horizon controller", func() {
                        })
                        keystoneAPI := keystone.CreateKeystoneAPI(namespace)
                        DeferCleanup(keystone.DeleteKeystoneAPI, keystoneAPI)
+                       th.SimulateDeploymentReadyWithPods(
+                               horizonName,
+                               map[string][]string{deploymentName.Namespace + "/storage": {"172.18.0.1"}},
+                       )
+
                })
                It("Check the resulting endpoints of the generated sub-CRs", func() {
-                       th.SimulateDeploymentReplicaReady(horizonName)
                        horizon := GetHorizon(horizonName)
                        Expect(horizon.Spec.NetworkAttachments).To(Equal([]string{"storage"}))
+                       Expect(nad).To(Equal(horizon.Status.NetworkAttachments))
                })
        })
 })
``` 
[1] https://github.com/openstack-k8s-operators/horizon-operator/blob/main/controllers/horizon_controller.go#L808
[2] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/networkattachment/networkattachment.go#L101

fmount (changes requested):

fmount (approved):
/lgtm

bshephar (approved):
Great! Lgtm

PR #372, opened 2024-10-11
Remove OLM webhooks when running 'make run-with-webhook'

abays:
When using `make run-with-webhook`, local versions of the operator and its webhooks are added to the cluster.  If the operator was previously installed via OLM, then there might be lingering webhooks from that installation.  We've previously been assuming that the user would manually remove them, but we should just get rid of them ourselves since those OLM webhooks need to be deleted anyhow for the local webhooks to function unimpeded.

### DIFF
--- a/hack/configure_local_webhook.sh
+++ b/hack/configure_local_webhook.sh
@@ -87,3 +87,5 @@ webhooks:
 EOF_CAT
 
 oc apply -n openstack -f ${TMPDIR}/patch_webhook_configurations.yaml
+
+oc patch "$(oc get csv -n openstack-operators -l operators.coreos.com/horizon-operator.openstack-operators -o name)" -n openstack-operators --type=json -p="[{'op': 'remove', 'path': '/spec/webhookdefinitions'}]" || true


hack/configure_local_webhook.sh
```diff
@@ -87,3 +87,5 @@ webhooks:
 EOF_CAT
 
 oc apply -n openstack -f ${TMPDIR}/patch_webhook_configurations.yaml
+
+oc patch "$(oc get csv -n openstack-operators -l operators.coreos.com/horizon-operator.openstack-operators -o name)" -n openstack-operators --type=json -p="[{'op': 'remove', 'path': '/spec/webhookdefinitions'}]" || true
```
rabi, line 91: We document to remove webhooks along with the controller-manager in other operators though, rather than patching the csv in this script. Maybe we make it consistent across operators.

rabi (commented):
/lgtm

hack/configure_local_webhook.sh
abays: Agreed.  That was my thinking: replicate the pattern in all operators.  But first I wanted to get your feedback on this initial form.  Thanks for the review!

abays: @rabi What would you think about adding another `oc patch` to `configure-local-webhooks.sh` to scale-down the CSV's deployment replicas to zero and then changing the comments here [1] to just mention the cleanup (since we will automatically be handling the scale-down and removal of OLM webhooks)?  i.e. something like:

```
# The "configure-local-webhooks.sh" script below will remove any OLM 
# webhooks for the operator and also scale its deployment replicas down
# to 0 so that the operator can run locally.
# Make sure to cleanup the webhook configuration for local testing 
# before deplying with OLM again, i.e.
# $oc delete -n openstack validatingwebhookconfiguration/vhorizon.kb.io
# $oc delete -n openstack mutatingwebhookconfiguration/mhorizon.kb.io
```

[1] https://github.com/openstack-k8s-operators/horizon-operator/blob/506b2848e56b01f4e4eb12b43a6510d7486486c1/Makefile#L356-L361

rabi: Yeah sounds good, if we're automating it, we should do that as well. For local webhook cleanup there is a script already and we can just mention that. https://github.com/openstack-k8s-operators/horizon-operator/blob/main/hack/clean_local_webhook.sh

PR #376, opened 2024-10-14
Remove OLM webhooks when running 'make run-with-webhook', part 2

abays:
Follow-up to https://github.com/openstack-k8s-operators/horizon-operator/pull/372 to improve comments and further automate removal of OLM components before running the operator locally


rabi (commented):
/lgtm

PR #393, opened 2024-11-29
kuttl: switch reports to JUnitXML instead of JSON

ashu-011:
JUnitXML is the de-facto standard format for test reporting and widely supported by all tools which post-process test results.

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -14,7 +14,7 @@
 
 apiVersion: kuttl.dev/v1beta1
 kind: TestSuite
-reportFormat: JSON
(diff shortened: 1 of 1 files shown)


bshephar:
Is there an Epic for this? I feel like this is something that should be implemented in all repos if we're going to do it. Having inconsistencies like this between projects could be quite frustrating.

If this is ultimately to support log ingestion into Polarion or something, then I can see the value. But lets make sure it's consistent.

ashu-011:
Closing this as https://github.com/openstack-k8s-operators/horizon-operator/pull/425 was merged

PR #396, opened 2024-12-11
[httpd] Use absolute url with schema in redirect rule

olliewalsh:
Use the endpoint in the redirect rule to ensure it references the correct base url and schema which can be different to the current schema when TLS is terminated at the route

Added horizonEndpoint template param with the full url, renamed the incorrectly named horizonEndpointUrl to horizonEndpointHost.

Jira: [OSPRH-12005](https://issues.redhat.com//browse/OSPRH-12005)

### DIFF
--- a/templates/horizon/config/local_settings.py
+++ b/templates/horizon/config/local_settings.py
@@ -64,18 +64,26 @@ DEBUG = False
 def get_pod_ip():
     import socket
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+    hostport = (
+        "{{ .horizonEndpointHost }}",
+        {{- if .isPublicHTTPS }}
+        443
+        {{- else }}
+        80
+        {{- end }}
+    )
     try:
-        s.connect(("{{ .horizonEndpointUrl }}", 80))
+        s.connect(hostport)
         return s.getsockname()[0]
     except socket.gaierror:
         s.close()
         s = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
-        s.connect(("{{ .horizonEndpointUrl }}", 80))
+        s.connect(hostport)
         return "[{}]".format(s.getsockname()[0])
     finally:
         s.close()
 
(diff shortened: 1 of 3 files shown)


templates/horizon/config/local_settings.py
```diff
@@ -65,17 +65,17 @@ def get_pod_ip():
     import socket
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     try:
-        s.connect(("{{ .horizonEndpointUrl }}", 80))
+        s.connect(("{{ .horizonEndpointHost }}", 80))
```
stuggi, line 68: in case of tls, is this still port 80?

olliewalsh: IIUC this doesn't matter, just opens a socket to $something to determine the pod IP.
I guess if there is a firewall blocking port 80 on the ingress controller it would block this though so might be best to use the correct port

bshephar: Yeah, correct. I just needed a way to determine the pods IP address so that we could add it to `ALLOWED_HOSTS`. The port should be ok, because port 80 is always open it just redirects to 443, and an open port of any kind is sufficient to pick up the IP address.

bshephar (approved):

templates/horizon/config/local_settings.py
olliewalsh: However in this specific case port 80 is firewalled (not specified where/how though) so it may be safest to use the correct port for the endpoint

olliewalsh: Done in the latest patch. Maybe in a follow-up should look into replacing this with  [downward-api volume](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/) or [env var](https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/) passing status.podIP instead

stuggi (approved):
/lgtm

PR #403, opened 2025-01-24
Set ENABLE_WATCHER env variable to yes if watcher service exists

amoralej:
A new watcher-operator is being created [1].

It is desired that the watcher-dashboard is enabled in the horizon container if the service is deployed.

This patch is setting the env variable ENABLE_WATCHER to `yes` if the keystone service "watcher" is present. Otherwise, it is set to default value `no`.

Related: [OSPRH-11277](https://issues.redhat.com/browse/OSPRH-11277)

### DIFF
--- a/controllers/horizon_controller.go
+++ b/controllers/horizon_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"fmt"
 	"net/url"
+	"slices"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -98,6 +99,7 @@ type HorizonReconciler struct {
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete;
 //+kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneapis,verbs=get;list;watch;
 //+kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneendpoints,verbs=get;list;watch;
+//+kubebuilder:rbac:groups=keystone.openstack.org,resources=keystoneservices,verbs=get;list;watch;
 //+kubebuilder:rbac:groups=memcached.openstack.org,resources=memcacheds,verbs=get;list;watch;
 
 // service account, role, rolebinding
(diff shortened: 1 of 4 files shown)


controllers/horizon_controller.go
```diff
@@ -700,6 +738,18 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
...
 	// Create ConfigMaps and Secrets - end
 
+	// Check if keystone service exists for watcher in the same namespace
+	enabledServices := make(map[string]string)
+	for _, service := range keystoneServicesWatch {
+		enabledServices[service] = "no"
+		keystoneService, err := keystonev1.GetKeystoneServiceWithName(ctx, helper, service, instance.Namespace)
+		if err == nil && keystoneService != nil {
+			enabledServices[service] = "yes"
+		}
```
bshephar, line 748: The error isn't being handled here though. Maybe something like:

```suggestion
		keystoneService, err := keystonev1.GetKeystoneServiceWithName(ctx, helper, service, instance.Namespace)
		if err != nil  {
			return ctrl.Result{}, err
		}
		
		enabledServices[service] = "no"
		if keystoneService != nil {
		    enabledServices[service] = "yes"
		}
```

bshephar (changes requested):

controllers/horizon_controller.go
amoralej: IIUC, with your proposed patch, if the watcher service is not found the reconcile will return error and keep retrying instead of go ahead and create the deployment. However, in that case we want it to go on with the expected status, so I'd say we should not handle the error case?

My doubt is if my double condition "err == nil && keystoneService != nil" is redundant as maybe just checking for err would be enough.

amoralej: I've implemented this in last version using `if err != nil && !k8s_errors.IsNotFound(err)` . I think that makes what we need it to do.

bshephar: Sure, that sounds good to me. As long as we're checking it and not continuing on and causing a harder to trace failure somewhere further down the line.

Thanks.

bshephar (approved):

PR #413, opened 2025-02-20
Stop running Horizon as root

bshephar:


bshephar:
recheck

bshephar:
Makes sense:
```
httpd: Could not open configuration file /etc/httpd/conf/httpd.conf: Permission denied
```

So, I think if we want to go down this route of removing excessive privileges for each of the applications that we're running. We can start by removing Kolla and directly mounting things where they need to be like I've done in this change. But then it becomes a permissions problem, so we would still need some kind of initialization that executes as root, changes permissions on each of the required files, then goes away and we're left with a non priv user running the application.

pkg/horizon/deployment.go
```diff
@@ -44,9 +44,11 @@ func Deployment(
 	enabledServices map[string]string,
 	topology *topologyv1.Topology,
 ) (*appsv1.Deployment, error) {
-	runAsUser := int64(0)
 
-	args := []string{"-c", ServiceCommand}
+	var runAsNonRoot bool = false
+	var runAsUserGroup int64 = 8443
+
+	args := []string{"-DFOREGROUND"}
```
bshephar, line 51: The `-DFOREGROUND` is the first arg passed to the command here. :) @fmount

fmount:
obsoleted by #445. @bshephar feel free to close this one if you're ok with #445

PR #420, opened 2025-03-05
Support Horizon custom dashboard

fmount:
Demo: https://asciinema.org/a/706974
Jira: https://issues.redhat.com/browse/OSPRH-5948


config/samples/custom-theme/kustomization.yaml
```diff
@@ -0,0 +1,12 @@
...
+- ../base/openstack
+
+patches:
+- path: custom_theme.yaml
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+configMapGenerator:
+- files:
+  - ./_11_rcue_theme.py
+  - ./rcue.tar.gz
```
fmount, line 11: I might need an additional iteration to add `rcue.tar.gz` sample that can be used by the users.

config/samples/custom-theme/README.md
```diff
@@ -0,0 +1,274 @@
...
+toggle functionality will not be available in the dashboard UI.
+
+
+The configuration file must be mounted in
+`/etc/openstack-dashboard/local_settings.d` within the Pod. The horizon
+bootstrap process uses this file to resolve the theme path during the `python
+manage.py collectstatic` operation.
+
+After creating the `_11_rcue_theme.py` file, you must provide the theme assets
+as a **tar.gz** tarball. Horizon will unpack this tarball in
```
fultonj, line 58: Do you want to change `Horizon will unpack` to `A script in the Horizon operator will unpack`?

config/samples/custom-theme/README.md
```diff
@@ -0,0 +1,274 @@
...
+`/usr/share/openstack-dashboard/openstack_dashboard/themes`
+
+## Procedure
+
+To use a custom theme, create a `_11_rcue_theme.py` file and configure the
+`AVAILABLE_THEMES` as follows:
+
+```python
+# override the CUSTOM_THEME_PATH variable with this settings snippet
+AVAILABLE_THEMES = [('rcue', 'Custom Theme', 'themes/rcue')]
```
fultonj, line 37: As I was reading this I wondered what the relative path `themes/rcue` would be expanded to. I found it with some quick googling "The path must be relative to the openstack_dashboard directory" so I assume it's going under `/usr/share/openstack-dashboard/openstack_dashboard/themes`.

Do you want to add:
```
See the [Horizon Theme Documentation](https://docs.openstack.org/horizon/latest/configuration/themes.html) for more details.
```

pkg/horizon/deployment.go
```diff
@@ -31,7 +31,7 @@ import (
 
 const (
 	// ServiceCommand is the command used to run Kolla and launch the initial Apache process
-	ServiceCommand           = "/usr/local/bin/kolla_start"
+	ServiceCommand           = "/usr/local/bin/kolla_theme_setup && /usr/local/bin/kolla_start"
```
fultonj, line 34: Nice trick to unpack the theme this way.

templates/horizon/bin/kolla_theme_setup
```diff
@@ -0,0 +1,25 @@
...
+TARGET_THEME_DIR=${2:-/usr/share/openstack-dashboard/openstack_dashboard/themes}
+TAR_OPTS="-xzvf"
+
+function unpack_theme {
+  local theme="$1"
+  if [[ "$theme" == *.tar.gz ]]; then
+      echo "Unpacking theme ..."
+      tar "${TAR_OPTS}" "${SRC_CUSTOM_THEME_DIR}/${theme}" -C "${TARGET_THEME_DIR}"
+  else
+      echo "$theme is not valid, ignoring ..."
```
fultonj, line 15: Maybe give a more specific error message to help the user if they didn't read the doc?
```
echo "$theme does not end in tar.gz, ignoring ..."
```

fultonj (commented):
This looks reasonable to me. I'll suggest a few things you could consider for small improvements but they're not blocking.

config/samples/custom-theme/README.md
fmount: sure, I'll change that in my next iteration

config/samples/custom-theme/README.md
fmount: Thank you John, added!

config/samples/custom-theme/README.md
fmount: Done

ashu-011 (approved):

deshipu (approved):

config/samples/custom-theme/README.md
```diff
@@ -0,0 +1,277 @@
...
+
+#### 2. Create ConfigMap with the all required files
+
+```bash
+$ oc create cm horizon-theme --from-file=custom.tar.gz --from-file=_11_custom_theme.py
+$ oc create cm horizon-theme-alt --from-file=custom_alt.tar.gz
+```
+
+> **Note**: It is convenient to have multiple ConfigMaps due to the size of the
+tarball that might exceed the ConfigMap allowed size.
```
bshephar, line 199: 1 MiB is pretty small. I haven't really looked into any custom Horizon themes to understand how big these tarballs could be. But do we expect any issues there? 

I guess it's not really a problem for us to validate in the Horizon Operator, since the ConfigMap would just fail to be created if it's over 1Mib in size. Rather than explicitly mentioning the size limitation, we can instead link to the k8s documentation about it:
https://kubernetes.io/docs/concepts/configuration/configmap/#motivation

bshephar (approved):

config/samples/custom-theme/README.md
fmount: right, but in general for simple customizations a tar.gz (binaryData) can be provided by a ConfigMap. This represents our immediate way to test it w/ extraMounts, but clearly we can put files in an NFS share or anything else that is meaningful. I'll defer to the UI team to test alternative solutions if customers require it, but it would be something we build on top of extraMounts (replacing the ConfigMap object w/ something else). In addition, for CI purposes, I think a simple configmap is sufficient to provide only a few files.

PR #431, opened 2025-03-14
Exit early if no openstack secret is provided

bshephar:
We depend on a valid openstack secret in order to reconcile Horizon, this change ensures that we exit the reconcile early if no secret is provided.

https://issues.redhat.com/browse/OSPRH-10075

### DIFF
--- a/controllers/horizon_controller.go
+++ b/controllers/horizon_controller.go
@@ -607,6 +607,19 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
 	//
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
 	//
+
+	// We can't reconcile without a valid OpenStack secret. Exit early here and update the status.conditions to
+	// inform users of the issue.
+	if instance.Spec.Secret == "" {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.InputReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.InputReadyWaitingMessage))
+
+		return ctrl.Result{}, fmt.Errorf("no openstack secret has been provided. Unable to reconcile")
+	}
+
 	ospSecret, hash, err := oko_secret.GetSecret(ctx, helper, instance.Spec.Secret, instance.Namespace)
 	if err != nil {
 		if k8s_errors.IsNotFound(err) {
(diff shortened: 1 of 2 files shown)


bshephar:
This actually shouldn't be necessary, the field is required:
https://github.com/openstack-k8s-operators/horizon-operator/blob/main/api/v1beta1/horizon_types.go#L87-L89

We might need to dig a little more into that bug to see how they ended up in that situation.

controllers/horizon_controller.go
```diff
@@ -607,6 +607,23 @@ func (r *HorizonReconciler) reconcileNormal(ctx context.Context, instance *horiz
 	//
 	// check for required OpenStack secret holding passwords for service/admin user and add hash to the vars map
 	//
+
+	// We can't reconcile without a valid OpenStack secret. Exit early here and update the status.conditions to
```
fmount, line 611: I'm wondering if having no `.Spec.Secret` is something we should also block at webhooks level, so we don't risk to even trigger a reconciliation loop if the input is not present. Anyway, I think exit early sounds reasonable if `.Spec.Secret` is not present.
I'm saying this because I assume you can't deploy a Horizon CR w/o that field, that is required from a kubebuilder perspective [1], but no one prevent you to update the CR and delete it. In that case the spec changes, and a reconciliation is triggered, but because this is an invalid action from a user input point of view, it should be blocked via `ValidateUpdate` [2]. Perhaps openstack-operator doesn't call that webhook, which is an additional issue I see here, and we should have a Jira to implement this missing part (we need both `HorizonSpecCore` Validation{Create,Update}, and we need to update openstack-operator to call them).

[1] https://github.com/openstack-k8s-operators/horizon-operator/blob/main/api/v1beta1/horizon_types.go#L89
[2] https://github.com/openstack-k8s-operators/horizon-operator/blob/main/api/v1beta1/horizon_webhook.go#L114

fmount: Done https://issues.redhat.com/browse/OSPRH-14726

fmount (approved):
/lgtm

controllers/horizon_controller.go
bshephar: Yeah, probably nice to block the CR creation to begin with. But yeah, if we do manage to be in this situation I just wanted to make sure we return a sane error via the `status.conditions`. 

I started looking into the webhooks today too after I made this comment:
https://github.com/openstack-k8s-operators/horizon-operator/pull/431#issuecomment-2723818535

PR #435, opened 2025-03-17
Ensure log file permissions for Apache

bshephar:
We use non root UID and GID's for starting the Apache process. Accordingly, any log files that it tries to create will need to have the same permissions to make them accessible to the Apache user. This change ensures all files under the default log directory (/var/log/horizon) have the current UID and GID set before trying to write to them.

Jira: https://issues.redhat.com/browse/OSPRH-9336

### DIFF
--- a/templates/horizon/config/local_settings.py
+++ b/templates/horizon/config/local_settings.py
@@ -29,7 +29,7 @@ LOGIN_REDIRECT_URL = '/dashboard/'
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 OPENSTACK_ENDPOINT_TYPE = "internalURL"
 OPENSTACK_API_VERSIONS = {
-  'identity': 3,
+    'identity': 3,
 }
 HORIZON_CONFIG["password_validator"] = {
     "regex": '',
@@ -42,13 +42,13 @@ DEBUG = False
 # This setting controls whether or not compression is enabled. Disabling
 # compression makes Horizon considerably slower, but makes it much easier
 # to debug JS and CSS changes
-#COMPRESS_ENABLED = not DEBUG
+# COMPRESS_ENABLED = not DEBUG
 
 # This setting controls whether compression happens on the fly, or offline
 # with `python manage.py compress`
 # See https://django-compressor.readthedocs.io/en/latest/usage/#offline-compression
 # for more information
-#COMPRESS_OFFLINE = not DEBUG
+# COMPRESS_OFFLINE = not DEBUG
 
 # If horizon is running in production (DEBUG is False), set this
 # with the list of host/domain names that the application can serve.
@@ -61,16 +61,18 @@ DEBUG = False
 # could land on any of the replicas. Instead, we need to explicity check the pod
 # we're currently running on. Therefore, we need to execute this function to
 # retrieve the IP address, which we will then in turn add to the ALLOWED_HOSTS list.
+
+
 def get_pod_ip():
     import socket
     s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
     hostport = (
         "{{ .horizonEndpointHost }}",
-        {{- if .isPublicHTTPS }}
+        {{- if .isPublicHTTPS}}
         443
         {{- else }}
         80
-        {{- end }}
+        {{- end}}
     )
     try:
         s.connect(hostport)
@@ -83,18 +85,19 @@ def get_pod_ip():
     finally:
         s.close()
 
+
 ALLOWED_HOSTS = [get_pod_ip(), "{{ .horizonEndpointHost }}"]
 
 USE_X_FORWARDED_HOST = True
 
-{{- if .isPublicHTTPS }}
+{{- if .isPublicHTTPS}}
 SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 SECURE_SSL_REDIRECT = True
 CSRF_COOKIE_SECURE = True
 SESSION_COOKIE_SECURE = True
 SESSION_COOKIE_HTTPONLY = True
 SECURE_HSTS_SECONDS = 31536000
-{{- end }}
+{{- end}}
 
 LOCAL_PATH = os.path.dirname(os.path.abspath(__file__))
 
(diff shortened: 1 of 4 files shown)


bshephar:
recheck

bshephar:
It doesn't like the patch command that I'm running:

```
=== CONT  kuttl/harness/basic-deployment
    harness.go:397: loading /go/src/github.com/openstack-k8s-operators/horizon-operator/tests/kuttl/tests/basic-deployment/03-customconfig-patch.yaml: error decoding yaml /go/src/github.com/openstack-k8s-operators/horizon-operator/tests/kuttl/tests/basic-deployment/03-customconfig-patch.yaml: error unmarshaling JSON: while decoding JSON: Object 'Kind' is missing in '{"op":"replace","spec":{"customServiceConfig":"# ANSSI\nOPERATION_LOG_ENABLE = True\nOPERATION_LOG_OTIONS = {\n  'mask_fields' : ['password', 'secret'],\n  'target_methods' : ['POST', 'GET', 'PUT', 'DELETE'],\n  'format' : (\"[%(domain_name)s] [%(domain_id)s] [%(project_name)s]\"\n  \" [%(project_id)s] [%(user_name)s] [%(user_id)s] [%(request_scheme)s]\"\n  \" [%(referer_url)s] [%(request_url)s] [%(message)s] [%(method)s]\"\n  \" [%(http_status)s] [%(param)s]\"),\n}\nLOGGING = {\n    'version': 1,\n    'disable_existing_loggers': False,\n    'formatters': {\n        'console': {\n            'format': '%(levelname)s %(name)s %(message)s'\n        },\n        'verbose': {\n            'format': '%(asctime)s %(process)d %(levelname)s %(name)s '\n                      '%(message)s'\n        },\n        'operation': {\n            # The format of \"%(message)s\" is defined by\n            # OPERATION_LOG_OPTIONS['format']\n            'format': '%(message)s'\n        },\n    },\n    'handlers': {\n        'null': {\n            'level': 'DEBUG',\n            'class': 'logging.NullHandler',\n        },\n        'console': {\n            # Set the level to \"DEBUG\" for verbose output logging.\n            'level': 'DEBUG' if DEBUG else 'INFO',\n            'class': 'logging.StreamHandler',\n            'formatter': 'console',\n        },\n        'operation': {\n          'level': 'INFO',\n          'class': 'logging.FileHandler',\n          'filename': '/var/log/horizon/operation.log',\n          'formatter': 'verbose',\n        },\n    },\n    'loggers': {\n        'horizon': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'horizon.operation_log': {\n            'handlers': ['operation'],\n            'level': 'INFO',\n            'propagate': False,\n        },\n        'openstack_dashboard': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'novaclient': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'cinderclient': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'keystoneauth': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'keystoneclient': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'glanceclient': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'neutronclient': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'swiftclient': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'oslo_policy': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'openstack_auth': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        'django': {\n            'handlers': ['console'],\n            'level': 'DEBUG',\n            'propagate': False,\n        },\n        # VariableDoesNotExist error in the debug level from django.template\n        # is VERY noisy and it is output even for valid cases,\n        # so set the default log level of django.template to INFO.\n        'django.template': {\n            'handlers': ['console'],\n            'level': 'INFO',\n            'propagate': False,\n        },\n        # Logging from django.db.backends is VERY verbose, send to null\n        # by default.\n        'django.db.backends': {\n            'handlers': ['null'],\n            'propagate': False,\n        },\n        'requests': {\n            'handlers': ['null'],\n            'propagate': False,\n        },\n        'urllib3': {\n            'handlers': ['null'],\n            'propagate': False,\n        },\n        'chardet.charsetprober': {\n            'handlers': ['null'],\n            'propagate': False,\n        },\n        'iso8601': {\n            'handlers': ['null'],\n            'propagate': False,\n        },\n        'scss': {\n            'handlers': ['null'],\n            'propagate': False,\n        },\n    },\n} \n"}}'

```

But that command seems to work fine locally. I'll need to reproduce the Kuttl tests locally and figure it out.

bshephar:
recheck

deshipu:
I *really* don't think we should be manipulating files inside a settings file. The determining of the hosts is already a stretch, but this is simply wrong.

bshephar:
> I _really_ don't think we should be manipulating files inside a settings file. The determining of the hosts is already a stretch, but this is simply wrong.

Yeah, fair enough. Happy for alternatives to solve the reported problem.  This is really just a workaround for the underlying problem which is running the container as root to begin with.

The alternative that I can see is the complete removal of the root user. Which would definitely be preferable. But there's some technical challenges that need to be overcome there as well.

1. Kolla wants to do a few things as root. We could overcome this by just mounting everything in the correct locations to begin with and do away with the stuff we're doing here: 
https://github.com/openstack-k8s-operators/horizon-operator/blob/main/templates/horizon/config/horizon.json#L3-L68

2. We can't bind ports 80 and 443 as a non-root user. So we will need to use non-standard ports which isn't really an issue, but it's notable.

3. Apache won't be able to read its configuration files in `/etc`. So we'll need to move those to a location that is accessible by the apache (or whatever non privileged user you want) user.

deshipu:
How about we put that workaround in Kolla for now? Since it's already manipulating files and permissions, it seems like it belongs there more?

PR #448, opened 2025-04-24
Add extraMounts envTests

fmount:
This patch adds `extraMounts` `envTests` to make sure they are properly propagated to `Horizon`. In addition it adopts the new `AssertVolumeMountPathExists` function provided by `lib-common/modules/common`.

Jira: https://issues.redhat.com/browse/OSPRH-15839


tests/functional/base_test.go
```diff
@@ -121,3 +121,37 @@ func GetSampleTopologySpec(label string) (map[string]interface{}, []corev1.Topol
 	}
 	return topologySpec, topologySpecObj
 }
+
+// GetExtraMounts - Utility function that simulates extraMounts pointing
+// to a  secret
+func GetExtraMounts(hemName string, hemPath string) []map[string]interface{} {
```
stuggi, line 127: if we use something like this in multiple operators, could add a test helper pkg to https://github.com/openstack-k8s-operators/lib-common/tree/main/modules/storage

stuggi (approved):
/lgtm

tests/functional/base_test.go
fmount: yeah, it's a good idea I agree

PR #485, opened 2025-07-28
add/setLimitRequestBody parameter to 10G

mcgonago:
Add/Set LimitRequestBody parameter to 10G

This is to get around the 1G limit [1]

[1] https://access.redhat.com/articles/6975397 

Jira: https://issues.redhat.com/browse/OSPRH-17359


stuggi:
If the idea is to set the `LimitRequestBody` to a fixed const, why not just add it to https://github.com/openstack-k8s-operators/horizon-operator/blob/main/templates/horizon/config/httpd.conf ?

I have rebased https://github.com/openstack-k8s-operators/horizon-operator/pull/402, please have a look, which would be adding the same approach to horizon to be able to customize the httpd vhost.

PR #489, opened 2025-08-04
Wrap func test with Eventually

stuggi:
This has been seen to fail in CI jobs. Wrapping it with Eventually should fix it.


abays (approved):
/lgtm

PR #506, opened 2025-09-01
Set HORIZON_IMAGES_UPLOAD_MODE to direct

fmount:
When not specified, `HORIZON_IMAGES_UPLOAD_MODE` has `legacy` as default value [0].
Because we already patched `glance-operator` to set the `cors` section automatically when horizon is enabled [1], this change allows horizon to rely on `direct` mode.

[0] https://docs.openstack.org/horizon/latest/configuration/settings.html#horizon-images-upload-mode
[1] https://github.com/openstack-k8s-operators/glance-operator/pull/787


fmount:
I will remove the  do-not-merge label if @deshipu approves it.

Issue #478, opened 2025-07-17
Horizon Not Starting in IPv6 SingleStack OpenShift Cluster


blackphantom39:
While trying to enable the deployment of Horizon in a Single Stack IPv6 cluster I came across the following error: 

```
ALLOWED_HOSTS = [get_pod_ip(), "horizon-openstack.apps.lab._redacted_"]
File "/usr/lib/python3.9/site-packages/openstack_dashboard/local/local_settings.py", line 72, in get_pod_ip
s.connect(hostport)
OSError: [Errno 101] Network is unreachable

```

From what I can tell the IPv6 supported added in #315 only works within DualStack setups, but I am not able to verify this. The problem to me seems to stem from the attempt in the try catch block to use IPv4 first:

```py
    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    try:
        s.connect(("{{ .horizonEndpointUrl }}", 80))
        return s.getsockname()[0]
    except socket.gaierror:
[...]
```
[Ref](https://github.com/hjensas/horizon-operator/blob/4e363362d14b51a39563ce89835b3ac611ba0e10/templates/horizon/config/local_settings.py#L66-L70)

As there would not be any IPv4 Network available it would raise the aforementioned OSError and never actually attempt to connect via IPv6. My crude attempt for a fix was to add a config map that overwrites the file with the `get_pod_ip()` function changed to the following:

```py
def get_pod_ip():
    import socket
    import errno

    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    hostport = ("{{ .horizonEndpointUrl }}", 443)
    try:
        s.connect(hostport)
        return s.getsockname()[0]

    except OSError as e:
        if e.errno == errno.ENETUNREACH:
            s.close()
            s = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
            s.connect(hostport)
            return "[{}]".format(s.getsockname()[0])
        else:
            raise

    except socket.gaierror:
        s.close()
        s = socket.socket(socket.AF_INET6, socket.SOCK_DGRAM)
        s.connect(hostport)
        return "[{}]".format(s.getsockname()[0])
    finally:
        s.close()
```

However, this resulted in other problems, as I assume the kolla config running on startup tries to change the file. If the lack of an IPv4 address is the actual problem here and the "fix" /workaround is acceptable I would submit a PR accordingly.

bshephar:
[...]

PR #550, opened 2026-01-07
Use lib-common DownwardAPI to retrieve pod IP

bshephar:
This change uses the `lib-common` env `DownwardAPI` [1] to write the pods IP to the env vars of the pod. We then retrieve this in `local_settings.py` to populate `ALLOWED_HOSTS`.

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/env/env.go#L100-L114

Fixes #487
 Jira: https://issues.redhat.com/browse/OSPRH-19216

### DIFF
--- a/internal/horizon/deployment.go
+++ b/internal/horizon/deployment.go
@@ -191,6 +191,7 @@ func getEnvVars(configHash string, enabledServices map[string]string) map[string
 	envVars["ENABLE_WATCHER"] = env.SetValue(enabledServices["watcher"])
 	envVars["CONFIG_HASH"] = env.SetValue(configHash)
 	envVars["UNPACK_THEME"] = env.SetValue("true")
+	envVars["POD_IP"] = env.DownwardAPI("status.podIP")
 
 	return envVars
 }
--- a/templates/horizon/config/local_settings.py
+++ b/templates/horizon/config/local_settings.py
@@ -56,35 +56,14 @@ DEBUG = False
 # For more information see:
 # https://docs.djangoproject.com/en/dev/ref/settings/#allowed-hosts
 
-# get_pod_ip retrieves the pod's primary interface IP address. This is necessary
-# due to the dynamic IP addressing of pods. The HealthCheck needs to be able to
-# check the specific pod. We can't simply check via the route, since such a check
-# could land on any of the replicas. Instead, we need to explicity check the pod
-# we're currently running on. Therefore, we need to execute this function to
-# retrieve the IP address, which we will then in turn add to the ALLOWED_HOSTS list.
-def get_pod_ip():
-    import socket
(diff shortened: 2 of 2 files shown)


bshephar:
@gibizer @fmount @stuggi Hey folks, happy New Year. Welcome to 2026. 

Trying to solve a problem with IPV6 addresses raised by:
https://github.com/openstack-k8s-operators/horizon-operator/issues/478

I had initially proposed doing it this way:
https://github.com/openstack-k8s-operators/horizon-operator/pull/496

I know that @deshipu wasn't a huge fan of that implementation from the beginning. So, this is an attempt to find a compromise and still fix the problem of allowing Liveness/Readiness probes to healthcheck the individual pods.

What do y'all think of this alternative?

fmount:
+1 from me as well. It is clean enough and the complexity to achieve the goal is minimal.

I wanted to check some logs coming at least from the build-deploy job, but I don't see horizon enabled [1]. 

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_horizon-operator/550/pull-ci-openstack-k8s-operators-horizon-operator-main-horizon-operator-build-deploy-kuttl/2009200602453643264/artifacts/horizon-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-6965f656a2712be18219bf47c45b31085032383eed3e09cbf2491ae5b2211ce0/namespaces/openstack/crs/openstackcontrolplanes.core.openstack.org/openstack-galera.yaml

gibizer (commented):
Happy New Year!

I only have clarification question but feel free to ignore me to merge this without me as I saw others are OK with this approach and I'm pretty unfamiliar with this.

I'm low on context here. Does simply adding "localhost" and "::1" to the ALLOWED_HOSTS not enough to allow the probes to access horizon?

bshephar:
> Happy New Year!
> 
> I only have clarification question but feel free to ignore me to merge this without me as I saw others are OK with this approach and I'm pretty unfamiliar with this.
> 
> I'm low on context here. Does simply adding "localhost" and "::1" to the ALLOWED_HOSTS not enough to allow the probes to access horizon? 

Hey!

Na, unfortunately not. But it is a fair question, so context for everyone else. ALLOWED_HOSTS is a list of sites Django is allowed to serve. So for example, horizon.example.net, 192.168.3.4, etc.

Ref:

https://docs.djangoproject.com/en/6.0/ref/settings/#allowed-hosts

So we need to add each pod IP to the list, otherwise the pod won't respond to the health check request.

PR #591, opened 2026-05-07
Add AGENTS.md

fmount:
This patch adds `AGENTS.md` to the repo to ensure any `AI agent` can follow the conventions defined for this operator. At the same time, it allows to not waste time navigating the repo to find initial information and save tokens.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,79 @@
+# AGENTS.md - horizon-operator
+
+## Project overview
+
+See [README.md](README.md) for a full project overview, configuration examples,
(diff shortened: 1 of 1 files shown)


mcgonago (approved):
Looks great, only a few suggestions for a possible follow up:
- Consider mentioning TLS config in `templates/` (significant for
  SSL-related work)
- Could list key Horizon CR spec fields for agents building new features
- Minor: `internal/webhook/v1beta1/` rather than `internal/webhook/`
  for path precision

LGTM

PR #602, opened 2026-06-19
Set ENABLE_CLOUDKITTY env variable to yes if cloudkitty service exists

jlarriba:
Similarly to the watcher dashboard, this sets the ENABLE_CLOUDKITTY env variable to "yes" when the keystone service "cloudkitty" is present, enabling the cloudkitty-dashboard in the horizon container. Otherwise, it defaults to "no".

### DIFF
--- a/internal/controller/horizon_controller.go
+++ b/internal/controller/horizon_controller.go
@@ -245,6 +245,7 @@ var allWatchFields = []string{
 }
 
 var keystoneServicesWatch = []string{
+	"cloudkitty",
 	"watcher",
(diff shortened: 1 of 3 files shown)


deshipu:
/approve

This makes sense, and the code looks consistent with what was done for watcher. I'm not sure about the tests, because I don't have the experience, but they seem to be passing. I haven't tested it on a deployment.

bshephar (approved):
lgtm.

stuggi (approved):

PR #604, opened 2026-06-24
(fr-6) Use dedicated startup probe with higher failure threshold

Deydra71:
The startup probe shared the same configuration as liveness/readiness probes, giving Horizon only ~40s to start. In resource-constrained environments startup exceeds this window causing CrashLoopBackOff.

Introduce `formatStartupProbe()` with FailureThreshold: 12 (allowing ~120s for startup), matching the pattern used by Cinder, Glance, and Manila operators.

Assisted-by: Claude Opus 4.6

Note: This is a repeated issue seen in SKMO CI runs that's deploying horizon in main region.

### DIFF
--- a/internal/horizon/deployment.go
+++ b/internal/horizon/deployment.go
@@ -68,7 +68,7 @@ func Deployment(
 
 	livenessProbe := formatProbes()
 	readinessProbe := formatProbes()
-	startupProbe := formatProbes()
+	startupProbe := formatStartupProbe()
 
 	envVars := getEnvVars(configHash, enabledServices)
 
@@ -201,6 +201,21 @@ func formatProbes() *corev1.Probe {
 		TimeoutSeconds:      5,
 		PeriodSeconds:       10,
 		InitialDelaySeconds: 10,
+		ProbeHandler: corev1.ProbeHandler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path: horizonDashboardURL,
+				Port: intstr.FromString(horizonContainerPortName),
+			},
+		},
+	}
+}
+
+func formatStartupProbe() *corev1.Probe {
+
+	return &corev1.Probe{
+		TimeoutSeconds:   5,
+		PeriodSeconds:    10,
+		FailureThreshold: 12,
 		ProbeHandler: corev1.ProbeHandler{
 			HTTPGet: &corev1.HTTPGetAction{
 				Path: horizonDashboardURL,


Deydra71:
recheck

internal/horizon/deployment.go
```diff
@@ -210,6 +210,21 @@ func formatProbes() *corev1.Probe {
 	}
 }
 
+func formatStartupProbe() *corev1.Probe {
+
+	return &corev1.Probe{
+		TimeoutSeconds:   5,
```
fmount, line 216: @Deydra71 o/ 
We recently introduced a new module [1] to provide an interface for probes. 
Using cinder here as just a simple example [2] that shows how to use the interface, you can basically create a `Probeset` struct with something like:

```golang
	apiProbes, err := probes.CreateProbeSet(
		int32(cinder.CinderPublicPort),
		&scheme,
		instance.Spec.Override.Probes,
		cinder.GetDefaultProbesAPI(timeout),
	)
``` 
or, for more advanced usage, like mariadb does [3], it is possible to also pass  a command and the handler type.
In general I think we could improve this code in main and add the probe interface as well, so we can take advantage of the override in case we need to tune this statefulset in production.
Let me know if that aligns with the goal of this patch, otherwise we can create a dedicated follow up that enhances the interface in main and then we backport to fr6 as well.



[1] github.com/openstack-k8s-operators/lib-common/modules/common/probes
[2] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/internal/cinderapi/statefuleset.go#L56C2-L62C3
[3] https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/internal/mariadb/statefulset.go#L138

Deydra71: I see, I overlooked the lib-common module :/ 

Looking at the examples I think it's pretty straightforward to update it in main, and then create backport only from the new one (and close this one).

I can work on it this week. @fmount Is there any Jira ttracking the implementation of `probes.OverrideSpec` across controlplane? So far I could find only support in storage operators and mariadb

stuggi: but keep in mind that this will be a CRD change. we can not just backport it and expect it to show up without an openstack-operator change. we need to plan for when it has to be released. otherwise we have to do an short term update and a longer term transition to this

fmount: Hi, 
> Is there any Jira ttracking the implementation of probes.OverrideSpec across controlplane? So far I could find only support in storage operators and mariadb

Because the interface has grown over the time, the idea is to add overrides only where we need, and I assume we can create stories under https://redhat.atlassian.net/browse/OSPRH-2490 to track the work.
I agree that we need to coordinate to make sure it will be available on a specific maintenance release and bump the openstack-operator (fr6) to get the CRD change as well.
For cinder we have an associated bug that will go out soon, not sure we want to take a similar approach, or we just close 2490 w/ FR6 and work on dedicated items (e.g. a new horizon bug that creates the bug-epic and the associated stream).

Deydra71: o/ @stuggi @fmount sorry about getting quiet. Given what you shared - is it okay to merge this hardcoded approach to FR6 branch, so the branches are in sync and as a follow up coordinate the CRD changes in horizon with another operators in planned release?

stuggi: can not speak to the values, but from the approach I think this sounds good, so we can work for fr7 for the crd change

fmount: +1 indeed, let's work out the plan we discussed for the FR7 timeframe, I'm good with the change as it is to solve the problem.

fmount (approved):
/lgtm as discussed, approving this patch to fix the immediate issue, we'll follow up w/ the probeset refactoring during the FR7 cycle.

Deydra71:
/test horizon-operator-build-deploy-kuttl

node failure seems to be resolved, this time test-phase succeeded, but must-gather phase failed, rerunning once more..

PR #605, opened 2026-06-24
Use crypto/rand for Django SECRET_KEY generation

stuggi:
Replace math/rand-based rand.String(10) (~47 bits entropy) with lib-common's util.GeneratePassword(50) which uses crypto/rand, matching Django's own default SECRET_KEY length.

The key is used to HMAC-sign CSRF tokens and session cookies.

Existing deployments are not affected — the secret is only generated when it does not already exist (ensureHorizonSecret checks for an existing secret before creating a new one).

Jira: [OSPRH-31813](https://redhat.atlassian.net/browse/OSPRH-31813)

### DIFF
--- a/internal/controller/horizon_controller.go
+++ b/internal/controller/horizon_controller.go
@@ -55,7 +55,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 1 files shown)


fmount (approved):
/lgtm makes sense to rely on [1]. Thanks @stuggi for this patch!

[1] https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/util/password.go#L27

PR #606, opened 2026-06-24
Use dedicated startup probe with higher failure threshold

Deydra71:
The startup probe shared the same configuration as liveness/readiness probes, giving Horizon only ~40s to start. In resource-constrained environments startup exceeds this window causing CrashLoopBackOff.

Introduce `formatStartupProbe()` with FailureThreshold: 12 (allowing ~120s for startup), matching the pattern used by Cinder, Glance, and Manila operators.

Assisted-by: Claude Opus 4.6

Note: This is a repeated issue seen in SKMO CI runs that's deploying horizon in main region.

### DIFF
--- a/internal/horizon/deployment.go
+++ b/internal/horizon/deployment.go
@@ -68,7 +68,7 @@ func Deployment(
 
 	livenessProbe := formatProbes()
 	readinessProbe := formatProbes()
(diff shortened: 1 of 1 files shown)


abays (approved):
/lgtm

PR #631, opened 2026-07-28
[18-stable] Remove unnecessary pods RBAC permissions

stuggi:
The workload rbacRules and the matching kubebuilder RBAC marker in the controller granted the workload service account full CRUD (create/delete/get/list/patch/update/watch) on core Pods, but that service account never reads or writes Pod objects directly. Remove the unused grants.

The kubebuilder RBAC marker for pods (get;list) is kept: lib-common's VerifyNetworkStatusFromAnnotation, called from this controller to verify NetworkAttachments, lists Pods using the controller-manager's own client, so the manager's ClusterRole (config/rbac/role.yaml) still needs get;list on pods. Regenerate config/rbac/role.yaml.

Includes ac2eccbc8fe4eee7435ec1a4e6b4bd1b454ed55b

(cherry picked from commit 45e0ef86a1e1b8f673c730ffd72c827a43be66cc)

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -8,7 +8,6 @@ rules:
   - ""
   resources:
   - configmaps
-  - pods
   - secrets
   - services
   verbs:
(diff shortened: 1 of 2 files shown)


abays (approved):
/lgtm

PR #445, opened 2025-04-16
Do not run horizon with root user

fmount:
This patch removes `root` from `horizon` `Pods`. The operator now builds pods using the `kolla` provided `user/group`. As a result of that an additional security context has been defined for the running container.

Jira: https://issues.redhat.com/browse/OSPRH-15373

PR #635, opened 2026-08-03
Harden Horizon workload SA and pod security context

HanzJas:
Remove hostmount-anyuid from the Horizon ServiceAccount SCC grant and stop auto-mounting the API token into the web-facing pod. Also set AllowPrivilegeEscalation to false and drop all capabilities so a compromise of the Horizon container cannot abuse hostPath pod creation or the kube API.

Keep anyuid so the workload can still run as the fixed Apache UID. ConfigMap, Secret, EmptyDir, and PVC ExtraMounts remain supported.

Assisted-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -122,14 +122,6 @@ rules:
   verbs:
   - use
 - apiGroups:
-  - security.openshift.io
-  resourceNames:
-  - hostmount-anyuid
-  resources:
-  - securitycontextconstraints
-  verbs:
-  - use
-- apiGroups:
   - topology.openstack.org
   resources:
   - topologies
--- a/config/samples/custom-theme/README.md
+++ b/config/samples/custom-theme/README.md
@@ -8,6 +8,14 @@ you can load custom themes using the
 [ExtraMounts](https://github.com/openstack-k8s-operators/dev-docs/blob/main/extra_mounts.md)
 feature.
 
+## ExtraMounts volume support
+
+To load custom themes, use a ConfigMap or a PVC.
+
+Inline NFS volumes (`volumes[].nfs`) and hostPath mounts are **not** supported.
+The Horizon ServiceAccount does not use the `hostmount-anyuid` SCC that
+OpenShift requires for these mount types.
+
 ## Theme configuration
 
 To provide additional themes, you must specify them in the `AVAILABLE_THEMES`
--- a/internal/controller/horizon_controller.go
+++ b/internal/controller/horizon_controller.go
@@ -118,7 +118,6 @@ type HorizonReconciler struct {
 // +kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=get;list;watch;create;update;patch
 // service account permissions that are needed to grant permission to the above
 // +kubebuilder:rbac:groups="security.openshift.io",resourceNames=anyuid,resources=securitycontextconstraints,verbs=use
-// +kubebuilder:rbac:groups="security.openshift.io",resourceNames=hostmount-anyuid,resources=securitycontextconstraints,verbs=use
 // +kubebuilder:rbac:groups="",resources=pods,verbs=get;list
 // +kubebuilder:rbac:groups=k8s.cni.cncf.io,resources=network-attachment-definitions,verbs=get;list;watch
 // service account, role, rolebinding
@@ -1222,7 +1221,7 @@ func configureHorizonRbac(ctx context.Context, helper *helper.Helper, instance *
 	rbacRules := []rbacv1.PolicyRule{
 		{
 			APIGroups:     []string{"security.openshift.io"},
-			ResourceNames: []string{"anyuid", "hostmount-anyuid"},
+			ResourceNames: []string{"anyuid"},
 			Resources:     []string{"securitycontextconstraints"},
 			Verbs:         []string{"use"},
 		},
--- a/internal/horizon/deployment.go
+++ b/internal/horizon/deployment.go
@@ -28,6 +28,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
 )
 
 const (
@@ -119,7 +120,8 @@ func Deployment(
 					Labels:      labels,
 				},
 				Spec: corev1.PodSpec{
-					ServiceAccountName: instance.RbacResourceName(),
+					ServiceAccountName:           instance.RbacResourceName(),
+					AutomountServiceAccountToken: ptr.To(false),
 					Containers: []corev1.Container{
 						// the first container in a pod is the default selected
 						// by oc log so define the log stream container first.
--- a/test/kuttl/tests/tls-deployment/01-assert.yaml
+++ b/test/kuttl/tests/tls-deployment/01-assert.yaml
@@ -90,6 +90,7 @@ spec:
           name: horizon-tls-certs
           readOnly: true
           subPath: tls.key
+      automountServiceAccountToken: false
       securityContext: {}
       serviceAccount: horizon-horizon
       serviceAccountName: horizon-horizon


fmount:
seems a valid failure [1] and the container does not reach the bootstrap part.

[1] https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openstack-k8s-operators_horizon-operator/635/pull-ci-openstack-k8s-operators-horizon-operator-main-horizon-operator-build-deploy-kuttl/2084308181986578432/artifacts/horizon-operator-build-deploy-kuttl/openstack-k8s-operators-gather/artifacts/must-gather/quay-io-openstack-k8s-operators-openstack-must-gather-sha256-2ad80528f9fada501ebd28e3022b02b06ba498b39da7f4f7061bc70d348212cd/namespaces/horizon-kuttl-tests/pods/horizon-6d96b55df9-2jqfn/horizon-6d96b55df9-2jqfn-describe

HanzJas:
Hello @fmount , thank you, I am still working on it.

But since you have shown up here, there is one thing that I wanted to discuss with someone more familiar with operators.
And it is `inline NFS`. Specifically this patch: https://github.com/openstack-k8s-operators/horizon-operator/pull/414
added _hostmount-anyuid for Horizon Service Account_ so users could use `inline NFS` volumes via `ExtraMounts` without setting `privileged: true`. In this patch is also mentioned _This can be removed once we are no longer running with UID 0._

This condition was met here,in your _non-root switch patch_, we have specific UID for Apache: https://github.com/openstack-k8s-operators/horizon-operator/pull/445, one problem is gone - we are no longer running with UID 0, but if I understand it correctly then without `hostmount-anyuid` (which we want to avoid from security reasons) using `inline NFS` will not work again because `NFS` and `hostPath` requires `hostmount-anyuid` as `anyuid` is enough only for `Default`/`ConfigMap`/`PVC ExtraMounts`.

1) Do we want to just remove `hostmount-anyuid` and accept that `inline NFS ExtraMounts` are unsupported?
2) Should we document somewhere that `PVC ExtraMounts` are possible but `inline NFS` not?
3) Is there any other way to allow `inline NFS` without granting `hostmount-anyuid`? Maybe some custom SCC?

fmount:
Hi @HanzJas,

> Hello @fmount , thank you, I am still working on it.
> 
> But since you have shown up here, there is one thing that I wanted to discuss with someone more familiar with operators. And it is `inline NFS`. Specifically this patch: #414 added _hostmount-anyuid for Horizon Service Account_ so users could use `inline NFS` volumes via `ExtraMounts` without setting `privileged: true`. In this patch is also mentioned _This can be removed once we are no longer running with UID 0._
> 
> This condition was met here,in your _non-root switch patch_, we have specific UID for Apache: #445, one problem is gone - we are no longer running with UID 0, but if I understand it correctly then without `hostmount-anyuid` (which we want to avoid from security reasons) using `inline NFS` will not work again because `NFS` and `hostPath` requires `hostmount-anyuid` as `anyuid` is enough only for `Default`/`ConfigMap`/`PVC ExtraMounts`.

Correct, that would be the main problem I guess. I didn't check all the potential `Volumes` we allow but `NFS` is definitely the right example.
> 
>     1. Do we want to just remove `hostmount-anyuid` and accept that `inline NFS ExtraMounts` are unsupported?

The main use case for `NFS` `ExtraMounts` on `Horizon` would be unpacking custom themes (rather than using a ConfigMap), which would work equally well through a PVC-backed NFS mount, so I think we could go this path and document that mounting through an `inline nfs` is not supported. I think this is a good tradeoff and allows to meet the security constraints we're trying to get.

> 
>     2. Should we document somewhere that `PVC ExtraMounts` are possible but `inline NFS` not?
Yes, we should update https://github.com/openstack-k8s-operators/horizon-operator/tree/main/config/samples/custom-theme to state what is supported/unsupported.
> 
>     3. Is there any other way to allow `inline NFS` without granting `hostmount-anyuid`? Maybe some custom SCC?
I think, as I mentioned in point 1, we can use a PVC-backed NFS, which is basically a regular PVC that has NFS at that level [1]. This would work at OCP level, and doesn't require extra privileges for horizon pod. you can still use extraMounts but I never tried to create a procedure with this use case. It's something we could explore in the future and POC that.
In the meantime, I would just put `inline NFS` as an unsupported path. 

[1] https://kubernetes.io/docs/concepts/storage/persistent-volumes/#persistent-volumes

HanzJas:
Hello @fmount , thank you, I did it the way you recommend, could you please review this PR when you have a minute?

fmount:
> Hello @fmount , thank you, I did it the way you recommend, could you please review this PR when you have a minute?

Hi @HanzJas thanks for pinging. I'm wondering if this patch is covered by the broader effort about removing kolla and fixing privileges and therefore obsoleted by #636 (which is a pattern we're spreading across the operators). 
It would be better to converge to that effort, so I'm wondering if we should close this one in favor of #636 and provide feedback on it (it would avoid unnecessary refactor/rebases).
What do you think?

HanzJas:
> > Hello @fmount , thank you, I did it the way you recommend, could you please review this PR when you have a minute?
> 
> Hi @HanzJas thanks for pinging. I'm wondering if this patch is covered by the broader effort about removing kolla and fixing privileges and therefore obsoleted by #636 (which is a pattern we're spreading across the operators). It would be better to converge to that effort, so I'm wondering if we should close this one in favor of #636 and provide feedback on it (it would avoid unnecessary refactor/rebases). What do you think?

Oh, okay, I did not know about this patch, it did not exist when I started to work on it.
I agree that it does not make much sense to have two parallel hardening patches (especially if it is a pattern you are spreading across all operators).
I can see that the patch from @stuggi  already sets `automountServiceAccountToken: false` so this part is definitely redundant.
But I can see an important difference in dropping `hostmount-anyuid` that we discussed together + documented PVC-only themes which is not included in the [#636](https://github.com/openstack-k8s-operators/horizon-operator/pull/636).
If I understand this PR then it brings better default hardening using `nonroot-v2` but `hostmount-anyuid` still kept NFS supported and does not close the possibility of post-RCE hostPath escalation reported here: https://redhat.atlassian.net/browse/OSPRH-33188

stuggi:
> > > Hello @fmount , thank you, I did it the way you recommend, could you please review this PR when you have a minute?
> > 
> > 
> > Hi @HanzJas thanks for pinging. I'm wondering if this patch is covered by the broader effort about removing kolla and fixing privileges and therefore obsoleted by #636 (which is a pattern we're spreading across the operators). It would be better to converge to that effort, so I'm wondering if we should close this one in favor of #636 and provide feedback on it (it would avoid unnecessary refactor/rebases). What do you think?
> 
> Oh, okay, I did not know about this patch, it did not exist when I started to work on it. I agree that it does not make much sense to have two parallel hardening patches (especially if it is a pattern you are spreading across all operators). I can see that the patch from @stuggi already sets `automountServiceAccountToken: false` so this part is definitely redundant. But I can see an important difference in dropping `hostmount-anyuid` that we discussed together + documented PVC-only themes which is not included in the [#636](https://github.com/openstack-k8s-operators/horizon-operator/pull/636). If I understand this PR then it brings better default hardening using `nonroot-v2` but `hostmount-anyuid` still kept NFS supported and does not close the possibility of post-RCE hostPath escalation reported here: https://redhat.atlassian.net/browse/OSPRH-33188

Thanks for the pointer. I have update the PR to remove `hostmount-anyuid`.

HanzJas:
> > > > Hello @fmount , thank you, I did it the way you recommend, could you please review this PR when you have a minute?
> > > 
> > > 
> > > Hi @HanzJas thanks for pinging. I'm wondering if this patch is covered by the broader effort about removing kolla and fixing privileges and therefore obsoleted by #636 (which is a pattern we're spreading across the operators). It would be better to converge to that effort, so I'm wondering if we should close this one in favor of #636 and provide feedback on it (it would avoid unnecessary refactor/rebases). What do you think?
> > 
> > 
> > Oh, okay, I did not know about this patch, it did not exist when I started to work on it. I agree that it does not make much sense to have two parallel hardening patches (especially if it is a pattern you are spreading across all operators). I can see that the patch from @stuggi already sets `automountServiceAccountToken: false` so this part is definitely redundant. But I can see an important difference in dropping `hostmount-anyuid` that we discussed together + documented PVC-only themes which is not included in the [#636](https://github.com/openstack-k8s-operators/horizon-operator/pull/636). If I understand this PR then it brings better default hardening using `nonroot-v2` but `hostmount-anyuid` still kept NFS supported and does not close the possibility of post-RCE hostPath escalation reported here: https://redhat.atlassian.net/browse/OSPRH-33188
> 
> Thanks for the pointer. I have update the PR to remove `hostmount-anyuid`.

Okay guys, as those changes will be covered in the [#636](https://github.com/openstack-k8s-operators/horizon-operator/pull/636) PR, I will close this one. Thank you!

PR #636, opened 2026-08-06
Remove kolla entrypoint pattern, harden SecurityContext

stuggi:
Replaces the kolla_start/kolla_set_configs/config.json staging pattern with direct volume mounts to final config paths and direct service commands, and hardens the pod's SecurityContext.

SecurityContext: horizon's kolla-era httpd/horizon worker processes always ran as apache (confirmed via `ps -ef` on a running kolla pod), never under a dedicated per-service system user the way keystone/manila do -- openstack-dashboard's RPM has never shipped its own user. Rather than migrate to serviceuser.HorizonUID like every other operator (which would silently break read/write access to any ExtraMount-backed content still owned apache:apache from before this migration), the pod keeps apache (serviceuser.ApacheUID) as its primary identity, with serviceuser.HorizonUID granted as a supplemental group instead. RBAC is additive
(hostmount-anyuid;nonroot-v2), since horizon's documented custom-theme ExtraMount mechanism can plausibly need NFS.

Custom theme mechanism: preserved via a non-root "theme-setup" init container that seeds a writable EmptyDir with the image's own baked-in themes plus any user-supplied *.tar.gz theme (via theme_setup, renamed from kolla_theme_setup since it was never actually an image-baked kolla script -- it's always lived in this repo).

.horizon-secret: re-materialized via a dedicated "secret-setup" init container into a pod-owned EmptyDir with real 0600 permissions, since Kubernetes always mounts Secret-backed files root-owned with at least group-read forced on, which horizon's own secret_key.read_from_file() rejects as insecure. Writes through a whole-directory mount (no SubPath) so the file exists as a real file before any later container references it via a SubPath mount of the same path -- a SubPath mount of a not-yet-existing path is otherwise auto-created by kubelet as a directory, permanently.

kolla_extend_start logic: brought in-repo as
templates/horizon/bin/dashboard_setup (a trimmed adaptation, not a runtime dependency on the image continuing to ship the original script), run by a "dashboard-setup" init container after theme-setup and secret-setup. Handles per-service dashboard panel enabling (Designate/Heat/Ironic/Manila/Octavia/Watcher/CloudKitty) and settings-changed-detection driving "manage.py collectstatic"/ "compress". Removed the dead heat_policy.json copy from config_heat_dashboard -- the image now ships heat_policy.yaml directly via RPM into /etc/openstack-dashboard/, so the script's .json reference was a no-op. Changed cp -a to cp -r in config_dashboard since ownership preservation silently fails as non-root.

Horizon has a unique 3-init-container pipeline: theme-setup seeds themes, secret-setup re-materializes the secret key with correct 0600 permissions, and dashboard-setup enables dashboard panels and regenerates static assets. All three must run in order.

Other key changes:
- run-httpd mount path: /etc/httpd/run -> /run/httpd (canonical)
- Config volume DefaultMode: 0640 -> 0440
- EmptyDir volumes: adopted lib-common volume.WritableDirVolume() helpers with standardized constants
- Fix TLS/mTLS staging-path bugs (final paths expected by httpd.conf but not supplied by lib-common's default helpers)
- Fix missing top-level httpd ErrorLog directive
- Fix STATIC_ROOT in local_settings.py (was unset, causing collectstatic to write to Django's package-relative default)
- AutomountServiceAccountToken: false
- Update functional and kuttl tests

Jira: [OSPRH-33504](https://redhat.atlassian.net/browse/OSPRH-33504)
Jira: [OSPRH-33503](https://redhat.atlassian.net/browse/OSPRH-33503)

Depends-On: openstack-k8s-operators/lib-common#728


stuggi:
recheck

stuggi:
recheck

stuggi:
recheck

internal/horizon/deployment.go
```diff
@@ -119,7 +210,62 @@ func Deployment(
...
+					// horizon's kolla-based httpd workers always ran as
+					// apache, never under a dedicated "horizon" system
+					// user (confirmed via `ps -ef` on a running kolla
+					// pod) -- so apache is used here as the primary
+					// identity, with HorizonUID granted as a
+					// supplemental group instead, to preserve
+					// read/write compatibility with any pre-existing
+					// ExtraMount content (NFS, pre-provisioned PVCs)
+					// still owned by apache from before this migration.
+					SecurityContext: pod.RestrictivePodSecurityContext(users.ApacheUID, users.HorizonGID),
```
abays: This mentions adding a supplemental group, but it's not passed as such.  Seems like it's the main group.

stuggi: right, let me update the comment

stuggi: done

abays (approved):
/lgtm

